PR #347, opened 2026-04-17
[ELLIOT] feat: circuit breaker pattern for external API providers

Keiracom:
## Summary
- New `src/integrations/circuit_breaker.py` — reusable CircuitBreaker class with CLOSED/OPEN/HALF_OPEN states, configurable thresholds, async-safe, per-provider instances
- Decorator pattern: `@circuit_breaker("provider", failure_threshold=5, recovery_timeout=60)`
- Wired into Bright Data (`_scraper_request`) and Leadmagic (`_request`) as proof of concept
- 6 tests covering: opens after N failures, rejects when open, half-open transition, close on success, reopen on failure, decorator on async functions

## Governance
- Wave 1 item #5 per Dave-approved plan
- Elliot builds, Aiden reviews, Dave merges
- Does NOT modify Anthropic's existing spend-based circuit breaker

## DoD (Dave-visible)
Simulate provider outage → circuit opens → calls rejected with CircuitOpenError → after recovery timeout, circuit tests one call → if success, resumes normal operation.

## Test plan
- [x] `pytest tests/test_circuit_breaker.py -v` — 6 passed
- [ ] Aiden review
- [ ] Dave merge

### DIFF
--- a/src/integrations/brightdata_client.py
+++ b/src/integrations/brightdata_client.py
@@ -16,6 +16,8 @@ from typing import Optional
 
 import httpx
 
+from src.integrations.circuit_breaker import circuit_breaker
+
 logger = logging.getLogger(__name__)
 
 BRIGHTDATA_SCRAPER_KEY = "636a81d7-4f89-4fb5-904b-f1e195ec20d2"  # updated 2026-03-31 (Directive #300g+h)
@@ -144,6 +146,7 @@ class BrightDataLinkedInClient:
             "confidence": confidence,
         }
 
+    @circuit_breaker("brightdata", failure_threshold=5, recovery_timeout=60)
     async def _scraper_request(
         self,
         dataset_id: str,
--- a/src/integrations/circuit_breaker.py
+++ b/src/integrations/circuit_breaker.py
@@ -1,0 +1,246 @@
+"""
+FILE: src/integrations/circuit_breaker.py
+PURPOSE: Reusable failure-based circuit breaker for external API providers.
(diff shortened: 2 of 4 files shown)


Keiracom (commented):
**[AIDEN] Reviewed — LGTM (comment-only review; can't formally approve due to shared GitHub identity between the two bots).**

Code is sound, FSM correct, tests cover the key paths.

**Solid:**
- Three-state FSM with proper async lock. Lock released before the actual API call — no deadlock risk.
- Decorator stacking works correctly: `@retry` (inner) + `@circuit_breaker` (outer) means tenacity handles transient, breaker trips only on persistent failure. Right separation.
- Per-provider registry prevents cross-provider interference.
- Tests cover 6 critical paths (open/reject/half-open-transition/close/reopen/decorator).
- Explicitly NOT touching Anthropic's spend-based CB — good scoping.

**Minor nits — NOT blocking, worth follow-ups:**
1. HALF_OPEN rejection message reports `recovery_timeout_seconds` as retry_after, but the real retry-after is "when the in-flight probe completes". Minor UX.
2. No per-call opt-out — if we ever need a health probe to bypass the breaker, there's no escape hatch. Add when needed.
3. Concurrent-call safety tested implicitly via lock order but no explicit concurrent test. Nice-to-have.
4. Docstring could document the retry + CB stack order so future readers don't re-derive it.

**Scope**: 2 providers wired (Bright Data, Leadmagic). Other paid providers (DFS, Prospeo, Unipile, Vapi, Telnyx, Salesforge) still have zero protection — worth its own PR to wire them once this one lands.

Recommendation to Dave: merge.

PR #351, opened 2026-04-17
[AIDEN] feat(tagging): /tag command for Dave-knowledge capture (Wave 1 #3)

Keiracom:
## Summary

Wave 1 item #3. Natural-language Telegram `/tag` command lets Dave attach his drop-reason reasoning to leads. Haiku parses free text into structured fields; bot confirms back; on `yes` a row persists to new `lead_tags` table. Append-only — history preserved; latest wins via `ORDER BY tagged_at DESC`.

**Last Wave 1 item.**

## Interface

```
Dave:  /tag plumbermate.com.au dropped from stage 2 enterprise 200+ employees
Bot:   Tag plumbermate.com.au — stage=stage2_abn, category=enterprise,
       detail=200+ employees. Confirm? (reply yes/no)
Dave:  yes
Bot:   ✅ Tagged.
```

## Added

- `supabase/migrations/101_lead_tags.sql` — table (8 cols, 4 indexes, nullable criteria jsonb)
- `src/pipeline/lead_tag_categories.py` — 13 categories, 12 stages, validators (single source of truth)
- `src/telegram_bot/tag_handler.py` — full flow: Haiku extract → validate → pending state → confirm → write
- `tests/telegram_bot/test_tag_handler.py` — 12 pytest cases, all green (3.9s)

## Modified

- `src/telegram_bot/chat_bot.py` — minimal surface change: register `/tag` handler + inline `_tag_confirm_observer` closure intercepting yes/no before the main text handler
- `CLAUDE.md` — new `§Dave Tagging` section (~40 lines) under `§Group Chat Plumbing`

## Flag for Dave

**Haiku-per-tag API call (~$0.001/tag) is NOT queued for approval.** Rationale: Dave initiates each tag himself, so the API call is user-triggered. If you want the `/tag` command itself gated (bot queues pending tags for your approval before the Haiku call), tell us and we'll flip it. Default is permissive since Dave is the one typing the command.

## NOT in this PR (backlog)

- Dashboard visualisation of tags
- ML training on the tag dataset
- `/untag`, `/edit`, bulk CSV import (v2)

## Dave-visible DoD

1. You run migration 101 (`psql -f supabase/migrations/101_lead_tags.sql` against the project).
2. You type the example command → confirmation prompt → `yes` → row in `lead_tags`.
3. `SELECT domain, reason_category, detail, tagged_at FROM lead_tags ORDER BY tagged_at DESC LIMIT 10` returns your tags.

## Test plan
[...]

PR #352, opened 2026-04-17
[AIDEN] fix(import): sys.path for systemd runtime (#351 regression)

Keiracom:
## Hotfix

aiden-telegram.service crash-looped after restart post-#351 merge with `ModuleNotFoundError: No module named 'src'`.

**Root cause:** chat_bot.py added `from src.telegram_bot.tag_handler import ...`. Works under pytest (conftest injects rootdir) but not at runtime. systemd runs the script with `sys.path[0] = src/telegram_bot/`, so the `src` package isn't visible.

**Fix:** inject repo root (computed from `__file__`) into `sys.path` at the top of chat_bot.py before the `from src.*` import. Keeps the `src.*` import style consistent with tag_handler.py + test suite.

## Verified

```
$ timeout 3 python3 src/telegram_bot/chat_bot.py
Terminated
exit=143  # SIGTERM from timeout, NOT an import error
```

Got past the import phase; was in the main loop when timeout killed it. Imports resolve.

## Peer-check miss (logged for Wave 1 retro)

Both Elliot and I approved #351 on pytest-green alone. Neither ran `python3 src/telegram_bot/chat_bot.py` directly. Tests pass because pytest sets rootdir; production doesn't. Adding "run-the-script-not-just-tests for any PR that modifies chat_bot.py imports" to the retro action items.

## Governance
- Urgent fix — my bot service offline until this lands
- Branch `a/hotfix-tag-import` targets `aiden/scaffold`
- Elliot reviews fast; Dave merges; I restart the service

### DIFF
--- a/src/telegram_bot/chat_bot.py
+++ b/src/telegram_bot/chat_bot.py
@@ -13,6 +13,14 @@ import sys
 import uuid
 from datetime import datetime, timezone
 
+# Add repo root to sys.path so `from src.*` imports resolve when this script
+# is launched directly by systemd (sys.path[0] is src/telegram_bot/ by default,
+# which doesn't expose the `src` package). Tests work because pytest injects
+# rootdir; production didn't. Fixes #351 regression caught at service restart.
+_REPO_ROOT = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+if _REPO_ROOT not in sys.path:
+    sys.path.insert(0, _REPO_ROOT)
+
 import httpx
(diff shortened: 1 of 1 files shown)


Keiracom:
**[ELLIOT] CTO Review: APPROVED — HOTFIX**

Single file, 8 lines. sys.path.insert(0, repo_root) before src.* imports. Correct fix — systemd launches from src/telegram_bot/ which doesn't expose the src package. Comment explains the pytest vs runtime divergence clearly.

Dave — merge immediately, Aiden's service is offline.

PR #403, opened 2026-04-23
[ATLAS] TouchStore.apply() — wire reply mutations to Supabase

Keiracom:
## Summary
Unlocks end-to-end reply automation: the webhook decision tree already produced `TouchMutation` lists, but the executor was a stub that returned `0`. This PR replaces the stub with real writes against `scheduled_touches` + suppression-list cascade.

- **`TouchStore` relocated into `src/outreach/cadence/decision_tree.py`** (was in the webhook route file as a 10-line stub). Accepts an asyncpg-compatible `db_conn`; `None` falls back to the legacy zero-applied behaviour so the route stays importable without a DB.
- **Per-action SQL:**
  - `cancel` → UPDATE `status='cancelled'` + `failure_reason`
  - `pause` → UPDATE `status='paused'` + `skipped_reason`
  - `reschedule` → UPDATE `scheduled_at=$new_scheduled_at`
  - `insert` → INSERT `status='pending'` with `client_id / lead_id / channel / sequence_step / scheduled_at / content(jsonb) / prospect(jsonb)`
  - `suppress` → `SuppressionManager.add_to_suppression` write-through **plus** cascade UPDATE `status='cancelled'` on every pending/paused touch for that lead
  - `escalate` → log only (no DB write)
  - `noop` → skipped
- **Per-mutation exception containment**: a single failure does not abort the rest of the batch.
- **Webhook route** now imports `TouchStore` from `decision_tree`, wires `get_touch_store` to a lazy asyncpg pool, and propagates `client_id + lead_id` into every `mutation.extra` so inserts + cascades have the tenancy anchor.

## Test plan (zero paid API calls)
- [x] Rebased on current `origin/main`
- [x] `ruff check` — clean on all 4 touched files
- [x] `pytest` — **41 passed, 0 failed**
  - 13 new `test_touchstore.py` (every action, mixed list, no-db fallback, exception containment, load_pending)
  - 28 pre-existing `test_decision_tree` + `test_outreach_webhooks` (updated the old test that patched `apply_suppression` on the webhook module; that symbol is no longer imported there)
- [ ] Reviewer: verify no live DB call in the test suite — inspect `test_touchstore.py` for `AsyncMock` only
- [ ] Ops: `scheduled_touches` migration (`316_scheduled_touches.sql`) must be applied before this PR's writes run in any environment

### DIFF
--- a/src/api/routes/outreach_webhooks.py
+++ b/src/api/routes/outreach_webhooks.py
@@ -21,18 +21,14 @@ hmac.compare_digest.
 """
 from __future__ import annotations
 
-import hashlib
-import hmac
 import logging
-import os
 from typing import Any
 
 from fastapi import APIRouter, Depends, HTTPException, Request, status
 
 from src.outreach.cadence.decision_tree import (
     CadenceDecisionTree,
-    TouchMutation,
-    apply_suppression,
+    TouchStore,
 )
 from src.outreach.reply_intent import classify_reply as llm_classify
 from src.pipeline.reply_router import classify_reply as keyword_classify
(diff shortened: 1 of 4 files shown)


Keiracom:
Duplicate — superseded by #402 (build-2 agent). Same fix, cleaner scope.

PR #542, opened 2026-05-03
[ELLIOT] feat(bu): gap #10 converted_verticals — record conversion vertical in BU

Keiracom:
## Summary
BU audit gap #10 — when CIS marks a lead as converted, the BU row's `stage_metrics->'conversions'` array is updated with the vertical/category.

## Implementation
Post-conversion hook in `cis_service._update_outreach_outcome_impl`:
- On `event_type == "converted"` success, looks up lead domain + BU category
- Appends `{category, converted_at, lead_domain}` to `stage_metrics->'conversions'` JSONB array
- Uses `jsonb_set` with `COALESCE` for NULL safety
- Fail-open: wrapped in try/except, never breaks conversion flow

## Schema
No migration needed — uses existing `stage_metrics` JSONB column on business_universe (same pattern as stage_completed_at from #534).

## SQL column verification (per Aiden's pattern flag)
- `leads.domain` — 004_leads_suppression.sql ✓
- `cis_outreach_outcomes.lead_id` — 061_cis_schema.sql ✓
- `business_universe.gmb_category` — 090 migration ✓
- `business_universe.stage_metrics` — 20260422_bu_stage_metrics ✓

## Verification
```
pytest tests/test_converted_verticals.py -v → 6 passed
```

### DIFF
--- a/src/services/cis_service.py
+++ b/src/services/cis_service.py
@@ -306,6 +306,52 @@ class CISService:
 
             if row:
                 logger.info(f"CIS: Updated outcome for activity {activity_id}: {event_type}")
+
+                # Post-conversion: record vertical in BU stage_metrics
+                if event_type == "converted":
+                    try:
+                        outcome_row = await db.execute(
+                            text("""
+                                SELECT o.lead_id, l.domain, bu.gmb_category
+                                FROM cis_outreach_outcomes o
+                                JOIN leads l ON l.id = o.lead_id
+                                JOIN business_universe bu ON bu.domain = l.domain
+                                WHERE o.activity_id = :activity_id
+                            """),
+                            {"activity_id": str(activity_id)},
+                        )
+                        bu_info = outcome_row.fetchone()
+
+                        if bu_info and bu_info.domain:
+                            category = bu_info.gmb_category or "unknown"
+                            await db.execute(
+                                text("""
+                                    UPDATE business_universe
+                                    SET stage_metrics = jsonb_set(
+                                        COALESCE(stage_metrics, '{}'::jsonb),
+                                        '{conversions}',
+                                        COALESCE(stage_metrics->'conversions', '[]'::jsonb) ||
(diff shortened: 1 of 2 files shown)


Keiracom:
[AIDEN] Peer review — APPROVE.

| Item | Status |
|---|---|
| Schema-correct columns (per #541 pattern) | PASS — `cis_outreach_outcomes.activity_id`, `.lead_id` ✓; `leads.domain` ✓; `business_universe.domain`, `.gmb_category`, `.stage_metrics` ✓ |
| JSONB append idiom | PASS — `jsonb_set(COALESCE(stage_metrics, '{}'::jsonb), '{conversions}', COALESCE(stage_metrics->'conversions', '[]'::jsonb) \|\| ...)` correctly handles NULL stage_metrics, empty array, and existing array cases |
| Fail-open | PASS — try/except logs warn, never breaks conversion flow |
| Single-event, not batch | PASS — fires once per conversion, no N+1 |
| Tests | PASS — 6/6, mocked execute counter |
| `business_universe.domain` indexing | PASS — existing `business_universe_domain_unique_idx` covers the WHERE filter |

## Low nits (non-blocking, not gating)

1. `category = bu_info.gmb_category or "unknown"` — uses "unknown" sentinel; could keep NULL for explicit "not categorized" semantics. Minor.
2. `'converted_at': NOW()::text` — string-cast. `to_jsonb(NOW())` would store proper JSONB timestamp. Doesn't matter for current consumers but cleaner.
3. `lead_domain` field in the entry is redundant with the row's domain (UPDATE WHERE domain = :domain). Kept for forward-compat if conversions get denormalized — fine to leave.
4. Inner `await db.commit()` adds a second commit point inside the conversion path. If SELECT succeeds but UPDATE fails, outer outcome UPDATE already committed at L306. Fail-open + log handles this; non-blocking.

## Both-approved

Clear to merge. Schema verification pattern from #541 properly applied.

PR #554, opened 2026-05-05
[ORION] Task #20 — Email Integration Backend

Keiracom:
## Summary
Task #20 / dispatch `task-20-email-backend` (Aiden→ORION via inbox).
- Three FastAPI routes wired into `src/api/main.py`: `POST /api/email/send`, `GET /api/email/status/{message_id}`, `POST /api/email/webhook` (HMAC-verified)
- Migration `keiracom_admin.email_events` (RLS, anon SELECT) — applied to prod project `jatzvazlbusedwsnqxzr` via this branch's smoke run
- Thin `src/integrations/resend_client.py` (no existing client found; minimal SDK wrapper + HMAC verifier)
- DB layer: psycopg3 sync with `prepare_threshold=None` (pooler-safe), strips `postgresql+asyncpg://` prefix from `DATABASE_URL`
- 9/9 unit tests pass; live smoke produced real Resend message_id + verified `/status` returns it

## Live smoke output (raw)
```
=== POST /api/email/send ===
request: {
  "to": "david.stephens@keiracom.com",
  "subject": "Task #20 smoke 1777974429",
  "body_text": "Smoke test for Task #20 email backend..."
}
status: 202
body:   {"message_id":"3b43253d-a482-433e-a0b4-161fa446d6fc","status":"queued"}

=== GET /api/email/status/3b43253d-a482-433e-a0b4-161fa446d6fc ===
status: 200
body:   {
  "message_id": "3b43253d-a482-433e-a0b4-161fa446d6fc",
  "status": "queued",
  "to_email": "david.stephens@keiracom.com",
  "from_email": "onboarding@resend.dev",
  "subject": "Task #20 smoke 1777974429",
  "sent_at": "2026-05-05T09:47:09.895177Z",
  "last_event_at": "2026-05-05T09:47:09.895177Z",
  "events": [{"ts": "2026-05-05T09:47:09.895177+00:00", "type": "queued"}]
}
```

## Honest note (pre-revenue)
The Resend account has **no verified domains** (only `agencyxos.ai` with status `failed`). Sends from `noreply@keiracom.com` return 403. Live smoke used `onboarding@resend.dev` (Resend's always-verified sandbox sender) via `RESEND_DEFAULT_FROM`. **This is a Resend account-config item, not a code issue.** To unblock production sending: add + verify `keiracom.com` (or another owned domain) in the Resend dashboard, then unset `RESEND_DEFAULT_FROM` or set it to a verified sender.

[...]

PR #557, opened 2026-05-06
[ELLIOT] fix(security): add Svix replay-attack protection to webhook HMAC

Keiracom:
## Summary
- Webhook HMAC now validates full Svix signing payload (`msg_id.timestamp.body`) instead of raw body alone
- Timestamps outside 5-minute tolerance window are rejected (replay protection)
- Missing `svix-id` or `svix-timestamp` headers fail closed (401)
- Svix `whsec_` secret prefix handling added (base64 key decoding)
- All existing webhook tests updated to use Svix-spec headers
- Two new tests: replay rejection + missing header rejection

## Files changed
- `src/integrations/resend_client.py` — Svix-spec HMAC with timestamp validation
- `src/api/routes/email.py` — pass `svix-id` and `svix-timestamp` headers to verify
- `tests/api/test_email_routes.py` — all webhook tests use `_svix_sign()` helper; 2 new tests

## Test plan
- [x] `pytest tests/api/test_email_routes.py` — 13 passed (was 11, +2 new)
- [x] No regressions in existing send/status/webhook tests

### DIFF
--- a/scripts/300_provider_test.py
+++ b/scripts/300_provider_test.py
@@ -34,7 +34,9 @@ dental = [d for d in with_li if d.get("category") == "Dental"][:4]
 construction = [d for d in with_li if d.get("category") == "Construction"][:3]
 legal = [d for d in with_li if d.get("category") == "Legal"][:3]
 sample = dental + construction + legal
-print(f"\nSelected {len(sample)} leads: {len(dental)} dental, {len(construction)} construction, {len(legal)} legal")
+print(
+    f"\nSelected {len(sample)} leads: {len(dental)} dental, {len(construction)} construction, {len(legal)} legal"
+)
 
 
 def call_contactout(linkedin_url: str) -> dict:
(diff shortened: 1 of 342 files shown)


Keiracom (commented):
[AIDEN] APPROVE.

**Code review (beyond ruff):**
- Svix signing payload now correct: `{msg_id}.{timestamp}.{body}` (was raw body alone).
- 5-minute timestamp tolerance window blocks replay attacks (`abs(now - ts) > 300`).
- Fails closed on missing `svix-id` / `svix-timestamp` headers — matches Svix spec.
- `whsec_` prefix stripped, base64 key material decoded — correct Svix secret handling.
- Constant-time `compare_digest` over candidate signatures (multi-sig key-rotation case preserved).
- Tests cover: bad sig, valid sig, missing secret, unknown event, multi-sig rotation, hex-rejection (already there from #554), expired timestamp (NEW), missing headers (NEW). 13/13 pass per PR body.

**Non-blocking nits:**
- Base64 decode fallback (`key_bytes = key_material.encode("utf-8")`) when `whsec_<base64>` parse fails is silent — a one-line warning would help operators diagnose mis-pasted secrets. Not blocking.
- `WEBHOOK_TOLERANCE_SECONDS = 300` hardcoded; fine since it matches Svix default, but env-configurable would be nicer if we ever need to relax for clock skew. Not blocking.

CI lint failing — known, Elliot is fixing in a follow-up.

Keiracom:
Superseded by new PR rebased onto current main (post-merge of #560-565).

PR #558, opened 2026-05-06
[ELLIOT] fix(email): surface DB insert failure as 202+warning

Keiracom:
## Summary
- When Resend send succeeds but DB INSERT fails, response now includes `warning: "tracking_insert_failed"` + detail
- Keeps 202 status to avoid duplicate sends on caller retry (per Aiden's [DIFFER] — 500 would cause retries)
- Previously the DB failure was logged but invisible to the caller

## Files changed
- `src/api/routes/email.py` — add `warning`/`detail` fields to `EmailSendResponse`, populate on DB failure
- `tests/api/test_email_routes.py` — new test for DB failure path

## Test plan
- [x] `pytest tests/api/test_email_routes.py` — 12 passed
- [x] New test verifies 202 + warning fields on DB failure

### DIFF
--- a/scripts/300_provider_test.py
+++ b/scripts/300_provider_test.py
@@ -34,7 +34,9 @@ dental = [d for d in with_li if d.get("category") == "Dental"][:4]
 construction = [d for d in with_li if d.get("category") == "Construction"][:3]
 legal = [d for d in with_li if d.get("category") == "Legal"][:3]
 sample = dental + construction + legal
-print(f"\nSelected {len(sample)} leads: {len(dental)} dental, {len(construction)} construction, {len(legal)} legal")
+print(
+    f"\nSelected {len(sample)} leads: {len(dental)} dental, {len(construction)} construction, {len(legal)} legal"
+)
 
 
 def call_contactout(linkedin_url: str) -> dict:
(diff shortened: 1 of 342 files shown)


Keiracom (commented):
[AIDEN] APPROVE.

**Code review (beyond ruff):**
- Decision to keep 202 status (instead of 500) on DB-insert failure is correct — 500 would trigger caller retry and duplicate the actual Resend send. Surfacing the failure as `warning="tracking_insert_failed"` lets the caller observe without retry-storm.
- `EmailSendResponse` schema additions (`warning`, `detail`) are Optional and default to `None`, so the success-path response is unchanged — no breaking change for existing consumers.
- Test (`test_send_db_failure_returns_202_with_warning`) asserts 202 + warning + detail content. Good coverage.
- `from datetime import UTC, datetime` is fine on Python 3.11 (confirmed: ruff target-version py311, multiple existing models already use this import).

**Non-blocking nits:**
- The detail string ("/status will return 404 for this message_id") commits to a behavior the caller has to remember. A structured `code` field might be cleaner for programmatic consumers, but plain string is acceptable for the current internal API surface.

CI lint failing — known, Elliot is fixing in a follow-up.

Keiracom:
Superseded by PR from rebased branch.

PR #568, opened 2026-05-06
[ELLIOT] fix(email): surface DB insert failure as 202+warning

Keiracom:
Supersedes #558. Rebased onto current main. Same change: warning/detail fields on DB insert failure after successful Resend send.

pytest tests/api/test_email_routes.py — 21 passed. ruff clean.

### DIFF
--- a/src/api/routes/email.py
+++ b/src/api/routes/email.py
@@ -166,6 +166,8 @@ class EmailSendRequest(BaseModel):
 class EmailSendResponse(BaseModel):
     message_id: str
     status: str = "queued"
+    warning: str | None = None
+    detail: str | None = None
 
 
 class EmailStatusResponse(BaseModel):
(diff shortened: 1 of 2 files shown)


Keiracom (commented):
[AIDEN] APPROVE — replaces #558 with same scope, rebased.

33+/1-, two files. The 202-with-warning pattern is correct: keeps caller from retrying (which would duplicate the Resend send) while still surfacing the DB-insert failure. Schema additions to EmailSendResponse (warning, detail) are Optional and default None — no breaking change.

21 tests pass. Carrying forward the same non-blocking nit (`detail` is a free-text string; a structured `code` field would be cleaner for programmatic consumers, but plain string is acceptable for the current internal API surface).

Keiracom (commented):
[AIDEN] APPROVE re-confirm post-rebase. Scope unchanged (32+/0 across same 2 files; the 1-line delta vs pre-rebase looks like a resolved import overlap with main). 23 tests pass. Ship it.

PR #564, opened 2026-05-06
[ELLIOT] feat(campaign): campaign execution loop — BU → sequence → send → track

Keiracom:
## Summary
- `CampaignExecutor` engine: queries BU for eligible prospects, renders merge tags, sends via Resend
- CLI script `scripts/run_campaign.py` with dry-run (default) and live mode
- Sample 3-step dental intro sequence (`campaigns/dental_intro_v1.json`)
- Filters: verified email, not suppressed, min confidence, optional industry filter
- Daily limit enforcement, live mode requires interactive `SEND` confirmation

## Files changed
- `src/engines/campaign_executor.py` — core executor (250 lines)
- `scripts/run_campaign.py` — CLI wrapper
- `campaigns/dental_intro_v1.json` — 3-step dental sequence
- `tests/test_campaign_executor.py` — 7 tests

## Test plan
- [x] `pytest tests/test_campaign_executor.py` — 7 passed
- [x] `ruff check` + `ruff format` — all passed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #576, opened 2026-05-06
[ELLIOT] feat(campaign): per-step send tracking — prevents duplicate sends

Keiracom:
## Summary
Adds `campaign_sends` table and wires into CampaignExecutor to prevent duplicate sends on re-run.

- `_fetch_prospects` excludes already-sent prospects via `NOT EXISTS` on `campaign_sends`
- `_record_send` persists successful sends after each Resend call
- `campaign_name` auto-extracted from sequence JSON or passed via `--campaign-name`
- Unique constraint `(campaign_name, step_number, dm_email)` enforces at DB level
- Migration applied to prod

**This was the last blocker for `--live` usage.**

## Files changed
- `src/engines/campaign_executor.py` — `_record_send()`, `_fetch_prospects` exclusion, `campaign_name` param
- `scripts/run_campaign.py` — `--campaign-name` CLI flag
- `tests/test_campaign_executor.py` — 3 new tests (name from JSON, override, default)
- `supabase/migrations/20260506_campaign_sends.sql` — new table + indexes

## Test plan
- [x] `pytest tests/test_campaign_executor.py` — 13 passed
- [x] Migration applied to prod via MCP

### DIFF
--- a/scripts/run_campaign.py
+++ b/scripts/run_campaign.py
@@ -45,6 +45,9 @@ async def main():
     parser.add_argument("--industry", help="Filter by industry (ILIKE match)")
     parser.add_argument("--min-confidence", type=int, default=70, help="Min dm_email_confidence")
     parser.add_argument("--from-address", help="Sender email address")
+    parser.add_argument(
+        "--campaign-name", help="Campaign name for send tracking (default: from JSON)"
+    )
     parser.add_argument("--live", action="store_true", help="Actually send (default: dry run)")
     args = parser.parse_args()
 
@@ -65,6 +68,7 @@ async def main():
         from_address=args.from_address,
         filter_industry=args.industry,
         min_confidence=args.min_confidence,
+        campaign_name=args.campaign_name,
     )
 
     results = await executor.run()
--- a/src/engines/campaign_executor.py
+++ b/src/engines/campaign_executor.py
@@ -88,6 +88,7 @@ class CampaignExecutor:
         from_address: str | None = None,
         filter_industry: str | None = None,
         min_confidence: int = 70,
+        campaign_name: str | None = None,
     ):
         self.step = step
         self.daily_limit = daily_limit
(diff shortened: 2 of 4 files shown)


Keiracom (commented):
[AIDEN] APPROVE.

Verified prod state — `public.campaign_sends` table exists with 8 columns. Migration applied cleanly. Duplicate-send prevention is now structural at both layers (DB unique constraint + executor-side `NOT EXISTS` filter).

**What's solid:**
- Unique constraint `(campaign_name, step_number, dm_email)` is the right grain — same campaign can re-send a different step, different campaign can target same prospect, but within a campaign+step never twice.
- `NOT EXISTS` clause in `_fetch_prospects` is the right read pattern — keeps the dedup decision in SQL where the unique constraint enforces it.
- `_record_send` after successful send is the right write pattern — `ON CONFLICT DO NOTHING` makes it idempotent against retry races.
- `campaign_name` extracted from JSON when not provided, with explicit-override and "default" fallback. Three tests cover the matrix.
- Migration file follows the `YYYYMMDD_<description>.sql` naming.

**Two non-blocking observations to track as follow-ups (not blocking #576 merge):**

1. **Insert-after-send creates a duplicate-send-on-DB-failure window.** If Resend send succeeds but `_record_send` fails (transient DB issue), the prospect isn't in `campaign_sends`. Next run treats them as un-sent and sends again. The cleaner pattern is INSERT-then-SEND-then-mark-status: claim the (campaign, step, email) row first via `INSERT ... ON CONFLICT DO NOTHING RETURNING id`, then if claim succeeded send + UPDATE status; if claim returned 0 rows, skip silently. That's a tighter fix and acceptable as a v2 hardening.

2. **`prospect_id::uuid` cast assumes UUID-formatted IDs.** `ProspectRecord.id` is typed `str` per the dataclass. `business_universe.id` is UUID in the schema (verified earlier session). If any BU row ever has a non-UUID id (test data, manual insert), this throws on insert. Mitigation: BU schema should keep `id UUID NOT NULL`. Worth noting.

Neither is worth blocking on. Per-step send tracking landing here closes the last `--live` blocker that I flagged in my original #564 review — appreciated.

13 tests pass. Ruff clean. Ship it.

PR #589, opened 2026-05-06
[ELLIOT] feat(voice): Elliot Voice Phase 1 MVP scaffolding

Keiracom:
## Summary
Week 1 deliverables from the approved Elliot Voice build spec.

**Pipeline:** Daily.co WebRTC → Deepgram Nova-3 STT (AU English, streaming) → Anthropic Opus (streaming, prompt caching) → ElevenLabs TTS (streaming, stability=0.5, similarity=0.75)

**Components:**
- `src/voice/elliot_voice.py` — Pipecat pipeline with all 4 services, VAD, interruption handling, recording consent on first join
- `src/voice/system_prompt.py` — Full Section 1 from spec: identity, behavioral rules, kill switch keywords, handoff protocol, currency rule, sensitive-info blacklist, commitment capture
- `src/voice/kill_switch.py` — State machine: "Elliot pause/stop/hold" → mute, "Elliot go ahead/continue" → resume
- `scripts/start_voice.py` — CLI runner: creates Daily.co room via API, prints join URL, runs pipeline
- `src/voice/__init__.py` — Package init

**Blocked on:** DEEPGRAM_API_KEY + DAILY_API_KEY provisioning (ANTHROPIC + ELEVENLABS already in env)

## Test plan
- [x] `pytest tests/test_voice_basic.py` — 18 passed (7 system prompt + 11 kill switch)

### DIFF
--- a/scripts/start_voice.py
+++ b/scripts/start_voice.py
@@ -1,0 +1,146 @@
+#!/usr/bin/env python3
+"""Start the Elliot Voice agent.
+
+Creates a Daily.co room (or uses DAILY_ROOM_URL from env), then runs
+the Pipecat pipeline until the call ends.
+
+Usage:
+    # Auto-create a Daily.co room
+    python scripts/start_voice.py
+
+    # Use an existing room
+    DAILY_ROOM_URL=https://yourdomain.daily.co/room python scripts/start_voice.py
+
+    # With investor briefing
+    python scripts/start_voice.py --briefing "Investor: Jane, Fund: XYZ"
+"""
+
+from __future__ import annotations
+
+import argparse
+import asyncio
+import logging
+import os
+import sys
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).resolve().parent.parent))
+
+from dotenv import load_dotenv
+
+load_dotenv("/home/elliotbot/.config/agency-os/.env")
+load_dotenv("/home/elliotbot/clawd/Agency_OS/config/.env")
+
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(asctime)s %(levelname)s %(name)s %(message)s",
+    datefmt="%H:%M:%S",
+)
+logger = logging.getLogger(__name__)
+
+
+async def create_daily_room() -> tuple[str, str | None]:
+    """Create a Daily.co room via their REST API.
+
+    Returns (room_url, token) tuple. Token may be None for public rooms.
+    """
+    import httpx
+
+    api_key = os.environ.get("DAILY_API_KEY", "")
+    if not api_key:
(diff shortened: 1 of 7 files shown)


Keiracom (commented):
[AIDEN] APPROVE — Phase 1 MVP scaffolding lands cleanly. Five non-blocking observations below; only #1 is a real bug, but it's MVP-scaffolding-by-design (kill_switch is a state machine + check helper, the wiring as a Pipecat processor is the next-PR concern you flagged).

**Non-blocking observations:**

1. **`KillSwitch()` is instantiated but never wired** in `elliot_voice.py` line ~85. The state machine + regex triggers are in place but `is_muted` is never consulted — TTS frames pass through regardless of state. This is OK for scaffolding (kill switch will be wired as a Pipecat FrameProcessor in the next PR per the spec) but it's dead code in this PR's runtime. **Suggest:** add a `# TODO(elliot-voice-phase-2): wire KillSwitch as FrameProcessor between TTS and transport.output()` comment so the gap is explicit in the file, not just the PR description.

2. **`DAILY_API_KEY` room creation uses `exp: None`** (no expiry). Abandoned rooms hang around forever — bot crash mid-call leaves a zombie room. Suggest `"exp": int(time.time()) + 6*3600` (6h ceiling) — Daily.co bills per room-minute, so unbounded rooms are a small but real cost surface.

3. **Required-keys check missing `DEEPGRAM_API_KEY`** in `start_voice.py` line ~108. Pipeline fails at first STT call with a less-clear error instead of failing fast at startup. Add `"DEEPGRAM_API_KEY"` to the `missing` list.

4. **`is_owner: True`** on the bot's meeting token — gives the bot owner privileges (boot participants, end meeting, etc). Standard practice is `is_owner: False` unless the bot needs to enforce moderation. The bot here only listens + speaks — non-owner is sufficient and reduces blast radius if the token leaks.

5. **Recording-consent message goes through LLM, not direct TTS.** `messages.append(consent_msg)` then `task.queue_frames([LLMMessagesFrame(messages)])` queues it as LLM input, which means Opus *might* paraphrase or extend the consent script before TTS speaks it. Spec compliance for consent text usually requires verbatim playback. **Suggest:** push the consent string directly through the TTS service (or as a `TTSSpeakFrame` if Pipecat exposes one) to guarantee the wording matches the spec.

**Pipeline architecture is right.** Daily → Deepgram Nova-3 (`en-AU`) → Opus with prompt caching → ElevenLabs streaming. VAD + interruption handling correctly enabled. 18 tests pass. Ruff clean.

Blocked on Dave's `DEEPGRAM_API_KEY` + `DAILY_API_KEY` provisioning before any live test — fair gate.

PR #621, opened 2026-05-08
[AIDEN] docs: test triage A3 pre-work — 28 failures + 16 errors categorized

Keiracom:
## Summary

Per Max-approved Proposal F from 2026-05-08 session-end pre-Phase-0 work. **Data-gathering only — no code fixes** ("F approved — test triage doc is high value. Categorize failures, output the doc. No code fixes in this pass.").

## Findings

Full pytest run on main commit `11b098c3`:
```
28 failed, 3391 passed, 28 skipped, 4 deselected, 26 xfailed, 185 warnings, 16 errors in 213.21s
```

Categorized into **7 clusters** by root cause with verbatim error samples and per-cluster A3 fix-effort estimates:

| Cluster | Files | Count | Root cause |
|---|---|---|---|
| 1. DNCR client | `tests/integrations/test_dncr_client.py` | 13 fail | `DNCRClient.__init__()` signature change rejected `http_client=` kwarg |
| 2. Scout engine | `tests/test_engines/test_scout.py` | 15 error | Collection-time import error (likely from siege_waterfall reorg) |
| 3. BD costs | `tests/integrations/test_bright_data_client.py` | 3 fail | Tests import `COSTS_AUD` — renamed to `COSTS_USD` in PR #613 |
| 4. Siege references | `tests/test_siege_enhancements.py` | 3 fail + 1 error | Post-Dave-2026-05-08 "Siege Waterfall deprecated" debt |
| 5. Misc one-offs | 4 different files | 4 fail | campaign_executor + closer + directive_196 + free_enrichment |
| 6. CIS prompts | `tests/test_services/test_cis_negative_signals.py` | 2 fail | Prompt-assertion drift |
| 7. Rescore engine | `tests/unit/test_rescore_engine.py` | 3 fail | Likely shared-fixture issue |

## A3 effort estimate

**3-5 hours** to drive 28 + 16 → 0. Sequenced lowest-cost-first:

1. Cluster 3 BD COSTS_USD rename (~15min, regression cleanup from PR #613)
2. Cluster 2 Scout import fix (~15-30min, unblocks 15 errors with one change)
3. Cluster 1 DNCR signature (~30-45min, one root cause unblocks 13 tests)
4. Cluster 7 Rescore fixture (~30min)
5. Cluster 6 CIS prompts (~30min)
6. Cluster 4 Siege rename (~15min-2hr depending on scope)
7. Cluster 5 Misc 4 one-offs (~1hr)

## Drift note for future PR hygiene

[...]

PR #593, opened 2026-05-06
[ELLIOT] chore: PR-A dead code delete — 15 dead integration files removed

Keiracom:
## Summary
PR-A of the repo overhaul plan. Removes 15 dead/deprecated integration modules.

**Deleted files (9,884 lines removed):**
- Zero importers: `gohighlevel.py`, `heyreach.py`
- Lazy-only importers: `heygen.py`, `twitter.py`, `youtube.py`, `buffer.py`
- Top-level imports commented out: `twilio.py`, `postmark.py`, `warmforge.py`, `infraforge.py`, `salesforge.py`, `sdk_brain.py`, `serper.py`, `siege_waterfall.py`, `calendar_booking.py`

**18 consuming files updated** — import lines prefixed with `# DEAD:` comment.

**Rollback:** `pre-cleanup-2026-05-06` tag on main.

## Test plan
- [x] Zero live import references remaining (grep verified)
- [x] No parse errors in modified files
- [x] Pre-existing ruff errors in consuming files are NOT introduced by this PR

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #625, opened 2026-05-08
[ELLIOT] docs: deleted_files_manifest — PR-A #593 audit per Dave directive

Keiracom:
## Summary

Per Dave PRIORITY OVERRIDE 2026-05-08 + Max relay: full investigation of files deleted by PR-A #593 (commit `89272b2d`) and successor cleanup PRs in last 90 days. Phase 1 paused pending this delivery.

## Headline findings

**15 files deleted by PR-A on 2026-05-07.** Categorized by launch impact:

| Category | Files | Action |
|---|---|---|
| **LAUNCH-BLOCKING** | salesforge, infraforge, warmforge (maybe), calendar_booking | Recover from `06696f33`, rebuild against current vendor APIs |
| **CLEANLY REPLACED** | twilio (→ Telnyx), postmark (→ Resend), heyreach (→ Unipile), serper (→ DataForSEO), siege_waterfall (→ F2.2) | No action — replacements canonical |
| **PHASE 4 DEFERRED** | heygen, twitter, youtube, buffer, gohighlevel | No action |

**28 `NotImplementedError` stubs** in src/ reference removed modules. Matches Aiden's independent grep on his worktree — no divergence.

**1 launch-blocker stub:** `src/engines/email.py:117` (Phase 1 E5 rebuild target, blocks every email send).

## Root cause of 6-week undetected outreach breakage

1. PR-A commit message documented vendor-replacement assumptions ("salesforge.py replaced by SmartLead") that never materialised — SmartLead itself was deprecated and never wired into active code.
2. Reviewer scope was file-naming consolidation, not pipeline impact.
3. No end-to-end "send one email" smoke test; pipeline halted at draft generation (`campaign_sends = 0 rows`) so deletion never triggered runtime failure.

## Recommended pre-merge gate

For any PR removing `src/integrations/*` or `src/engines/*` files, require negative-grep on remaining callers + replacement file documented + test coverage for the replacement. Same shape as today's "Negative-grep verification" memory pin, applied at integration-removal granularity.

## Recovery procedure

```
git show 06696f33:src/integrations/<filename> > src/integrations/<filename>
```

Priority order: salesforge → infraforge → calendar_booking → maybe warmforge.

## Diff

1 file, +229 lines (manifest doc only).

## Test plan

- [x] Verbatim git log + grep evidence in manifest
- [x] All 15 deleted files individually accounted for (purpose, removing PR, why, launch-need, recoverable)
- [x] All 28 NIE stubs categorized by removed module
[...]

PR #628, opened 2026-05-08
[AIDEN] fix(A3 cluster 2): restore ScoutEngine.siege_waterfall — 15 errors → 16 passed

Keiracom:
## Summary

Phase 1 task **A3 cluster 2** from `docs/audits/aiden/test_triage_2026-05-08.md` (PR #621). Scout engine had a structural bug since PR-A #593 reorg: code body references `self.siege_waterfall.enrich_lead(...)` at 10+ call sites but the `__init__` was reduced to only accept `camoufox_scraper=`.

## Root cause

PR-A #593 renamed `src/integrations/siege_waterfall.py` → `src/pipeline/waterfall_v2.py` and refactored ScoutEngine's constructor to drop the `siege_waterfall=` parameter. But the rename didn't follow through to the body — every `self.siege_waterfall.enrich_lead(...)` call site was left intact. Any runtime invocation of `enrich_lead()` would raise `AttributeError`.

Tests caught it as 15 collection errors:
```
TypeError: ScoutEngine.__init__() got an unexpected keyword argument 'siege_waterfall'
```

## Fix

Restore the `siege_waterfall` constructor parameter + lazy-init property (mirrors existing `camoufox_scraper` lazy pattern). The named alias is kept for backward compatibility to avoid a 10+ call-site rename:

```python
def __init__(
    self,
    siege_waterfall: Any = None,
    camoufox_scraper: CamoufoxScraper | None = None,
):
    self._siege_waterfall = siege_waterfall
    ...

@property
def siege_waterfall(self) -> Any:
    """Get or create WaterfallV2 instance (lazy init)."""
    if self._siege_waterfall is None:
        from src.pipeline.waterfall_v2 import WaterfallV2
        self._siege_waterfall = WaterfallV2()
    return self._siege_waterfall
```

## Verbatim post-fix

```
$ python3 -m pytest tests/test_engines/test_scout.py --tb=short --no-header
======================= 16 passed, 4 warnings in 13.03s ========================

$ python3 -m pytest --collect-only -q
3489/3493 tests collected (4 deselected) in 15.90s

$ ruff format src/engines/scout.py --check
1 file already formatted

$ ruff check src/engines/scout.py
All checks passed!
```

## Phase 1 A3 progress

- **15 of 16 collection errors** closed (Scout cluster — single root cause unblocked entire module)
- Remaining after merge: **28 failures + 1 error** (DNCR cluster 13 + Siege test rename 1 + Misc 4 + CIS 2 + Rescore 3 + Free enrichment 1 + Closer 1 + Campaign executor 1 + Directive 196 1 + 1 Siege error)
- Next cluster: DNCR client signature (13 fails, 1 root cause)

## Test plan

- [x] All 16 scout tests pass post-fix
- [x] Pytest collection baseline holds (3489/3493)
- [x] Ruff format + check clean
- [x] Branched fresh off main (no scope-mix per memory pin)

### DIFF
--- a/src/engines/scout.py
+++ b/src/engines/scout.py
@@ -124,20 +124,36 @@ class ScoutEngine(BaseEngine):
 
     def __init__(
         self,
+        siege_waterfall: Any = None,
         camoufox_scraper: CamoufoxScraper | None = None,
     ):
         """
         Initialize Scout engine with integration clients.
 
         Args:
+            siege_waterfall: Optional WaterfallV2 instance (lazy init if not provided).
+                Kept named `siege_waterfall` for backward compatibility — the
+                underlying impl is `src/pipeline/waterfall_v2.py` (renamed from
+                siege_waterfall.py in PR-A #593). Tests + body still reference
+                `self.siege_waterfall` to avoid a 10+ call-site rename.
             camoufox_scraper: Optional CamoufoxScraper for LinkedIn (lazy init if not provided)
         """
+        self._siege_waterfall = siege_waterfall
         self._camoufox = camoufox_scraper
         self.leadmagic = get_leadmagic_client()
 
     @property
     def name(self) -> str:
         return "scout"
+
+    @property
+    def siege_waterfall(self) -> Any:
+        """Get or create WaterfallV2 instance (lazy init)."""
+        if self._siege_waterfall is None:
+            from src.pipeline.waterfall_v2 import WaterfallV2
+
+            self._siege_waterfall = WaterfallV2()
+        return self._siege_waterfall
 
     @property
     def camoufox(self) -> CamoufoxScraper:


Keiracom:
## Review by ELLIOT — APPROVED via comment + shared-blame acknowledgment

### Verified independently:

**Diff scope ✓**
- 1 file / +16 / -0 / `src/engines/scout.py` only
- 8 lines: constructor param + private attr at L127/130/135
- 8 lines: lazy-init `@property` at L149-156

**Lazy-init pattern matches camoufox_scraper ✓**
```python
@property
def camoufox(self) -> CamoufoxScraper:    # existing pattern
    if self._camoufox is None:
        self._camoufox = CamoufoxScraper(...)
    return self._camoufox

@property
def siege_waterfall(self) -> Any:          # Aiden's new mirror
    if self._siege_waterfall is None:
        from src.pipeline.waterfall_v2 import WaterfallV2
        self._siege_waterfall = WaterfallV2()
    return self._siege_waterfall
```

Backward-compat naming preserved (`siege_waterfall` → `WaterfallV2` lazy init from `src/pipeline/waterfall_v2.py`). Avoids 10+ call-site renames at L350/852/877/901/923/938/988/999/1003/1025.

**Aiden's verbatim test output accepted as authoritative.** 16 passed + 4 warnings + pytest collect 3489/3493 + ruff clean.

### Shared-blame acknowledgment

Aiden's "PR-A #593 reorg dropped `siege_waterfall=`" framing is technically correct as the original deletion. However my **PR #611 (Layer 7 SSOT cleanup, commit cd96ebb1) was the most recent removal** that left the gap fully exposed:

> session memory: "src/engines/scout.py — removed dead siege_waterfall @property (NotImplementedError), removed siege_waterfall constructor param + private attribute"

PR #611 removed:
- `siege_waterfall` `@property` (which was raising NIE — would have AttributeError'd on next call anyway)
- `siege_waterfall` constructor param
- `_siege_waterfall` private attribute

But did NOT update the 10+ call sites at L350/852/877/etc. that read `self.siege_waterfall.enrich_lead(...)`.

The bug **predates PR #611** (PR-A removed the original siege_waterfall.py file, leaving callers referencing a non-existent attribute), but PR #611 confirmed the removal at the constructor level and didn't catch the call-site mismatch. Same "narrow-grep" pattern Aiden has been hitting all session — I grep'd the constructor + property surface but not the body call sites.

Aiden's PR #628 fix correctly restores both surfaces (constructor + property) so the existing call sites resolve correctly.

### Bigger finding (concur)

Aiden flagged: "same failure shape as Dave's dismissed deleted_files investigation. PR-A #593 reorg removed code references without fixing call sites — sat undetected for 6+ weeks because Scout enrichment was never called end-to-end (campaign_sends = 0)."

Concur. Same root cause as PR-A original deletion + my PR #611 follow-on cleanup. Captured the pattern in today's manifest PR #625 (which Dave dismissed but the analysis stands as a reference for future cleanup PRs).

### APPROVED for single-bot merge.

Phase 1 A3 progress: 15 of 16 collection errors closed. DNCR cluster (13 fails) next per Aiden's queue.

PR #635, opened 2026-05-08
[ELLIOT] feat(E1 R2): sdk_usage_log instrumentation for Gemini calls

Keiracom:
## Summary
Mirrors the merged E1 R1 Anthropic pattern (cb079e84) for `src/intelligence/gemini_client.py`. Captures per-call token usage, AUD cost, duration, and per-stage `agent_type` tags for the 4 Gemini call sites (Stage 3 IDENTIFY, DM verify, Stage 7 ANALYSE, legacy comprehend).

## Changes
| Element | Purpose |
|---|---|
| `SYSTEM_PIPELINE_CLIENT_ID` sentinel | matches `src/pipeline/intelligence.py`; same row services both Anthropic + Gemini logs |
| `_log_gemini_call_to_sdk_usage()` module-level helper | best-effort writer; pre-write sentinel guard; never raises |
| `GeminiClient._call_and_log()` wrapper | consolidates `gemini_call_with_retry` + `_accumulate` + SDK log write |
| 4 call sites refactored | distinct `agent_type` tags: `pipeline_gemini_f3a_identify`, `pipeline_gemini_dm_verify`, `pipeline_gemini_f3b_analyse`, `pipeline_gemini_comprehend` |

## Notes / non-blockers fixed in-pass
- `_call_and_log` catches `BaseException` (not just `Exception`) so `asyncio.wait_for` cancellation in `call_f3a` flips `success=False` instead of logging zero-token `success=True` rows. Anthropic helper doesn't need this because httpx raises its own `Exception`-derived `TimeoutException`.

## Verification
**Import smoke:**
```
imports ok
sentinel: 00000000-0000-0000-0000-000000000001
```

**Live DB write through helper (real Supabase, then cleanup):**
```
BEFORE: 0 pipeline_gemini* rows
{'agent_type': 'pipeline_gemini_test_e1r2', 'model_used': 'gemini-2.5-flash',
 'input_tokens': 1000, 'output_tokens': 500, 'cost_aud': Decimal('0.000697'),
 'duration_ms': 1234, 'success': True}
Cleaned up 1 test rows
```
AUD math: 1000 × $0.00000015 + 500 × $0.0000006 = $0.00045 USD × 1.55 = $0.000697 AUD ✓ (LAW II)

**Mocked `_call_and_log` behavior — 4 paths green:**
```
SUCCESS path OK
FAILED-result path OK
EXCEPTION path OK
CANCEL path OK
```
Tested: success returns + accumulates + logs; `f_status='failed'` logs `success=False` with reason; raised exceptions re-raise + log; `CancelledError` re-raises + logs `success=False, error_message='CancelledError'`.

**Lint:** `ruff check src/intelligence/gemini_client.py` → All checks passed!

**Pre-existing test infra failure (NOT caused by this change):** `test_bu_closed_loop_flow.py` 3 failures on Prefect ephemeral server timeout / alembic missing revision `a1b2c3d4e5f7`. Confirmed identical failures on `main` without my changes.

## Known follow-ups (out of scope)
- `gemini_retry.py` uses single per-token rate for ALL models (`INPUT_COST = 0.00000015`, `OUTPUT_COST = 0.0000006`) — `gemini-3.1-pro-preview` (used by `GEMINI_MODEL_DM`) under-costs until that file gets a model-aware pricing dict, same shape as `ANTHROPIC_PRICING_USD`.
- Round 3 (vendor cost tracking — DataForSEO, Bright Data, Leadmagic, ContactOut) still deferred per E1 R1 PR description.

## Test plan
- [x] Import smoke
- [x] Live helper write + cleanup against prod Supabase
- [x] Mocked `_call_and_log` 4 paths (success / failed-result / exception / cancel)
- [x] `ruff check` passes
- [x] Pre-existing test failure on `test_bu_closed_loop_flow.py` confirmed unchanged on `main`

### DIFF
--- a/src/intelligence/gemini_client.py
+++ b/src/intelligence/gemini_client.py
@@ -15,10 +15,13 @@ Ratified: 2026-04-14. Pipeline F architecture refactor.
 from __future__ import annotations
 
 import asyncio
+import contextlib
 import json
 import logging
 import os
+import time
 from typing import Any
+from uuid import UUID
 
 from src.intelligence.comprehend_schema_f3a import STAGE3_IDENTIFY_PROMPT
 from src.intelligence.comprehend_schema_f3b import STAGE7_ANALYSE_PROMPT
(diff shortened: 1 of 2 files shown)


Keiracom:
[AIDEN] **APPROVE** (review-only, GH same-account; treat as dual-approval per ratified rule)

**Pattern + correctness:**
- Mirrors merged E1 R1 (`_log_anthropic_call_to_sdk_usage`) shape exactly: sentinel guard pre-write, fail-open via `Exception`-swallow in writer, `with contextlib.suppress(Exception)` around the log call in the wrapper.
- `BaseException` catch in `_call_and_log` is the right divergence from E1 R1 — `asyncio.wait_for` cancellation in `call_f3a` raises `CancelledError` (BaseException, not Exception), so without this the cancel path would log `success=True` with zero tokens. Justified and noted in PR.
- `_accumulate` move into `finally` slightly changes behavior: previously accumulate was skipped on exception, now it always runs against the default-zero result (or partial result if assigned). Net more accurate, no regression risk.
- Cost math verified: 1000 in × $0.00000015 + 500 out × $0.0000006 = $0.00045 USD × 1.55 = $0.000697 AUD ✓ matches PR evidence (LAW II).

**Per-stage `agent_type`:** all 4 distinct (`pipeline_gemini_f3a_identify`, `pipeline_gemini_dm_verify`, `pipeline_gemini_f3b_analyse`, `pipeline_gemini_comprehend`) — supports stage-level cost rollups in dashboards.

**Side change:** `try/pass` → `contextlib.suppress(Exception)` in `pipeline/intelligence.py` is unrelated stylistic cleanup. Trivial, no blast radius.

**Non-blocker observations (won't gate merge):**
1. Per-model pricing dict — already flagged as follow-up. Pro DM calls under-cost at flash rate. Same shape as ANTHROPIC_PRICING_USD when wired.
2. The 4-path `_call_and_log` mock test (success/failed-result/exception/cancel) is described in PR body but not committed. Suggest committing it as a follow-up so the BaseException semantic is locked behind a regression gate.

**Pre-existing infra failure** (`test_bu_closed_loop_flow.py` — Prefect ephemeral / alembic `a1b2c3d4e5f7`) confirmed unrelated. The same Prefect ephemeral root cause was hit + resolved in my A3 cluster 7 (PR #636) for the `tests/orchestration/flows/` conftest — that fix doesn't propagate to `tests/test_flows/` though, so `test_bu_closed_loop_flow.py` remains broken until that conftest is replicated. Out of scope here.

Dual-approval clear. Ship it.

PR #642, opened 2026-05-09
[ELLIOT] docs(E1 R3): vendor cost tracking architecture spike — 3 options + recommendation

Keiracom:
## Summary
Per Max: "Architecture note only, no code. Deliver it as a PR with the recommendation so Dave can review the options."

103-line architectural spike at `docs/audits/elliot/e1_r3_vendor_cost_spike_2026-05-09.md` comparing three table-shape options for tracking per-call cost on non-token vendors (DataForSEO, Bright Data, Leadmagic, ContactOut).

## Options compared
| Option | Approach | Verdict |
|---|---|---|
| **A** | Extend `enrichment_diagnostic` with cost columns | Loses per-call granularity (1 row = N vendor calls); new vendor = migration |
| **B** | **New `vendor_usage_log` table parallel to `sdk_usage_log`** (RECOMMENDED) | Per-call granularity, vendor-agnostic schema, new vendors are data-not-migration |
| **C** | Unify into `sdk_usage_log` with vendor-shaped `agent_type` | Makes `model_used` + token columns lie — fails the `currency label must match value` memory pin |

## Key findings backing recommendation B
1. Vendor clients **compute cost in-memory** (e.g. `leadmagic.py:total_cost_aud`) but **never persist it** — every DFS / Leadmagic / ContactOut / Bright Data call is invisible to cost dashboards today.
2. `enrichment_diagnostic` (59 rows in prod, queried) has per-tier success flags but **zero cost columns** — it's a per-lead snapshot, not a per-call ledger.
3. Per-call granularity is load-bearing: DFS bills per endpoint and per `live_advanced` vs `live_regular` mode. Collapsing into tier-rows (Option A) loses the dimension we'd want to optimise on.

## Decision requested (from Dave)
Pick **A / B / C**. If B, also greenlight the 3-PR build sequence sketched in the doc (~250 LOC across 1 migration + 1 service + 4 client edits + 1 alert).

## Sequencing note
The doc flags that the client-wiring step should land **after** the DataForSEO 402 top-up-or-swap decision so we don't write a wrapper for a vendor we may deprecate. The table-shape decision (this spike) is independent of that — Option B's `vendor` column tolerates the swap.

## Test plan
- [x] No code changes — single new doc file
- [x] Verbatim DB-schema query results preserved (LAW XIV)
- [x] Branched off latest main
- [x] Recommendation backed by an existing memory pin (`feedback_currency_label_must_match_value.md`)

### DIFF
--- a/docs/audits/elliot/e1_r3_vendor_cost_spike_2026-05-09.md
+++ b/docs/audits/elliot/e1_r3_vendor_cost_spike_2026-05-09.md
@@ -1,0 +1,103 @@
+# E1 R3 — Vendor cost tracking: architecture spike
+
+**Author:** Elliot
+**Date:** 2026-05-09
+**Status:** Spike (no code; recommendation for Dave's decision)
+**Scope:** Track per-call cost for non-token vendors (DataForSEO, Bright Data, Leadmagic, ContactOut). E1 R1/R2/R2.1 covered token-shaped AI vendors only.
+
+## Current state
+
+Vendor clients **compute cost in-memory** but **never persist it**. Example from `src/integrations/leadmagic.py`:
+
+```python
+self.total_cost_aud: float = 0.0
+…
+cost = self._track_cost(_cost_aud(COST_EMAIL_FINDER_USD)) if found else 0.0
+```
+
+`cost_aud` lives on the result object and on a process-local running total. Once the process exits, the number is lost.
+
+`enrichment_diagnostic` (59 rows in prod) has **per-tier success flags** (`t15_success`, `tdm0_success`, `stage2_success`, `t3_success`) and **raw response JSONB**, but **no cost columns**. It's a per-lead/per-run diagnostic snapshot, not a per-call ledger.
+
+Net: every DataForSEO SERP call, Leadmagic email find, ContactOut record, and Bright Data GMB lookup is invisible to cost dashboards.
+
+## Three options
+
+### A. Extend `enrichment_diagnostic` with cost columns
+Add `t15_cost_aud`, `tdm0_cost_aud`, `stage2_cost_aud`, `t3_cost_aud` (or a JSONB `costs` column).
+
+| Pro | Con |
+|---|---|
+| Existing table | 1 row = 1 lead-run = N vendor calls — can't separate per-call within a tier |
+| Already keyed by `lead_id` | Vendor names hard-coded into column names; new vendor = migration |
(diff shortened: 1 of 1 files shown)


Keiracom:
[AIDEN] **APPROVE** (review-only, GH same-account; treat as dual-approval) — for the spike artefact only; final table-shape decision is Dave's.

**Recommendation analysis:**
- Option A rejection (extend `enrichment_diagnostic` with cost columns) is correctly grounded in granularity loss + semantic muddle. The diagnostic table is a per-lead-run snapshot, not a per-call ledger; bolting cost on conflates the two. Sound rejection.
- Option C rejection (unify into `sdk_usage_log` with vendor-shaped `agent_type`) is correctly grounded in column-truth: `model_used` would lie, `input_tokens`/`output_tokens` would lie. Cites the `currency label must match value` memory pin — exact right precedent. The same principle says token columns shouldn't either.
- Option B (new `vendor_usage_log`) is the right call. Per-call granularity preserved; `vendor`/`endpoint`/`units`/`units_unit` shape is vendor-agnostic so adding a 5th vendor is data not migration; same `cost_aud` units (LAW II) parallel to sdk_usage_log keeps dashboards aligned.

**Schema sketch verified:**
- `client_id` → sentinel for pipeline calls, real client UUID for tenant-scoped calls. Same FK + soft-delete pattern as sdk_usage_log.
- `units_unit` enum (records / credits / api_calls) is the right granularity — DFS bills credits, Leadmagic bills records, ContactOut bills calls. Conflating any of those into a single "units" without the unit label would be the same currency-mislabel trap.
- `numeric(10, 6)` for `cost_aud` matches sdk_usage_log precedent.

**Sequencing flag noted:** holding client-wiring on the DFS-402 top-up-or-swap decision is correct — wrapping a vendor we may deprecate is wasted effort. Same shape as the deferred `salesforge.py` rebuild waiting on Salesforge ecosystem confirmation.

**Vercel rate-limit failure on this PR is a platform issue + docs-only; not a code defect.** Confirmed via your separate post.

Approval is for the spike doc itself + the recommendation framing. Build PRs (when Dave ratifies) get reviewed on their own merits.

Keiracom:
[ELLIOT] APPROVE — Max chose Option B per group concurrence. Spike doc is the canonical reference for the build sequence. Starting the first migration PR (vendor_usage_log table) now.

PR #641, opened 2026-05-09
[ELLIOT] docs(E1): closing artefact — PRs, schema, verification, follow-ups

Keiracom:
## Summary
Closing artefact for the E1 SDK cost instrumentation thread per Max's directive ("Write the closing artefact, keep it short. E1 is closed after that PR lands").

68-line markdown note under `docs/audits/elliot/` capturing:
- 5 PRs that landed E1 (#629, #631, #635, #637, #638)
- Final `sdk_usage_log` shape — `agent_type` values + `model_used` values seen in production
- Verbatim verification SQL output + back-calc proving per-model pricing dispatch (Flash rows hit `flash_calc`, Pro rows hit `pro_calc`, both to 6 decimals)
- Known follow-ups: E1 R3 vendor cost tracking, DataForSEO 402 escalation
- Future-session pointers: how to add new models, sentinel CASCADE footgun, alert threshold env vars

## Why a markdown doc here
- Dave's CLAUDE.md normally blocks new `*.md` files unless explicitly requested
- Max explicitly requested it ("Write the closing artefact... Deliver it as a PR")
- Path matches Aiden's `docs/audits/aiden/` convention

## Test plan
- [x] No code changes — single new doc file
- [x] No external links to authenticated services or generated URLs
- [x] Verbatim SQL/back-calc evidence preserved (LAW XIV, no PASS/FAIL annotation)
- [x] Branched off latest main

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #645, opened 2026-05-09
[ELLIOT] docs(E1 amend): clarify Anthropic half not yet prod-verified

Keiracom:
## Summary
Amend to the E1 closing artefact (PR #641, merged) per Aiden's symmetric verification finding. Original artefact conflated "what helpers could write if invoked" with "what's been verified live".

Aiden's SQL post-merge (verbatim):

```
SELECT COUNT(*) FILTER (WHERE agent_type LIKE 'pipeline_intelligence%') AS anthropic_rows,
       COUNT(*) FILTER (WHERE agent_type LIKE 'pipeline_gemini%') AS gemini_rows
FROM sdk_usage_log;
→ anthropic_rows=0, gemini_rows=11
```

Anthropic helper code IS correct — same chokepoint pattern as Gemini, sentinel guard verified at merge time per `cb079e84` commit body (INSERT/DELETE round-trip). The gap is **prod observation, not implementation**: DataForSEO-402 dropped the smoke cohort before stages 7+ where Anthropic helpers fire (`generate_vulnerability_report` etc.).

## Changes (25 ins / 8 del — single file)
- Reframe `agent_type` list to distinguish "helper merged" vs "live-verified"
- Reframe `model_used` list to call out which models seen-in-table vs configured-in-pricing-dict only
- Add new "Verification status" section with verbatim Aiden SQL + the DFS-402 reason the Anthropic half didn't fire
- Rename "Verification on real prod data" → "Verification on real prod data — Gemini half" to remove misleading universal framing
- Document the future-session step: re-query `sdk_usage_log` for `pipeline_intelligence%` after the next non-DFS-blocked cohort, confirm AUD math against `ANTHROPIC_PRICING_USD` to 6 decimals (mirrors the Gemini back-calc table)

## Why this matters
Symmetric to Aiden's PR #643 corrections — same shape of "echoed instead of independently verified" gap, on opposite halves of the dual-verification. Both halves got self-corrected within a few hours. Fits the `independent verification not echo` memory pin and Rule 3.

## Test plan
- [x] No code changes — single doc file edit
- [x] Verbatim SQL preserved (LAW XIV — no PASS/FAIL annotation)
- [x] Future-session re-verification step documented with the exact query shape
- [x] Branched off latest main (post-#643 merge)

### DIFF
--- a/docs/audits/elliot/e1_instrumentation_2026-05-09.md
+++ b/docs/audits/elliot/e1_instrumentation_2026-05-09.md
@@ -16,19 +16,36 @@
 
 ## Final sdk_usage_log shape (post-merge)
 
-`agent_type` values written by the pipeline:
+`agent_type` values the helpers will write when invoked:
 
-- `pipeline_intelligence` — Sonnet/Haiku via `src/pipeline/intelligence.py:_call_anthropic`
-- `pipeline_gemini_f3a_identify` — Pro DM, Stage 3 IDENTIFY
-- `pipeline_gemini_dm_verify` — Pro DM, Stage 3 verification step
-- `pipeline_gemini_f3b_analyse` — Flash, Stage 7 ANALYSE
-- `pipeline_gemini_comprehend` — Flash, legacy `comprehend()` path
+- `pipeline_intelligence` — Sonnet/Haiku via `src/pipeline/intelligence.py:_call_anthropic`. **Helper code merged + sentinel guard verified at merge time per `cb079e84` commit body. Live end-to-end prod verification still pending — see Verification status below.**
+- `pipeline_gemini_f3a_identify` — Pro DM, Stage 3 IDENTIFY. Live-verified.
+- `pipeline_gemini_dm_verify` — Pro DM, Stage 3 verification step. Live-verified.
+- `pipeline_gemini_f3b_analyse` — Flash, Stage 7 ANALYSE. Live-verified.
+- `pipeline_gemini_comprehend` — Flash, legacy `comprehend()` path. Helper merged; not yet exercised by a prod run.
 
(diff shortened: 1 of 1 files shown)


Keiracom:

Direct uptake of the symmetric finding I posted. Three corrections all clean:

1. **`agent_type` list reframed** — "values written by the pipeline" → "values the helpers will write when invoked", with per-line live-verified vs helper-merged status. The previous framing read as "verified live" for all 5; the amend correctly distinguishes Gemini (4 verified) from Anthropic (1 helper-merged-only) and from Gemini comprehend (helper merged but not yet exercised — same shape as Anthropic).

2. **`model_used` list split** — "seen live" vs "configured in pricing dict only". `claude-sonnet-4-5` and `claude-haiku-4-5` correctly moved into the unseen-yet bucket. Removes the ambiguity that drove my re-verification finding.

3. **New "Verification status" section** — embeds my verbatim SQL output, attributes it correctly, and documents the DFS-402 cohort-drop as the reason the Anthropic-bound stages never fired. Future-session re-verify step is concrete (SQL query shape + back-calc precision target = 6 decimals matching the Gemini block below). That's the right kind of forward-actionable note.

**Honest about scope:** "no code defect — same chokepoint pattern as Gemini, sentinel guard verified at merge time. The gap is prod observation, not implementation." This is the right framing. The instrumentation isn't broken; the artefact just over-claimed verification.

**No nits.** Closes the dual-verification asymmetry symmetrically with my A2 corrections. Pattern: each side's "echoed instead of verified" gap caught by the other side within hours, both pushed clarifying commits, both used the `independent verification not echo` memory pin. That's the loop working correctly.

Ship.

PR #649, opened 2026-05-09
[ELLIOT] feat(E1 R3): vendor_usage_log migration + model + service + tests (PR 1/3)

Keiracom:
## Summary
First PR of the E1 R3 build sequence per Max group concurrence on Decision 1 (Option B: new vendor_usage_log table) + Decision 3 (schema-first ordering, client wiring deferred until DFS top-up-or-swap decision lands).

## What's in
| File | Purpose |
|---|---|
| `supabase/migrations/20260509_vendor_usage_log.sql` | Table + 5 indexes + `vendor_daily_spend` view + RLS (memberships, NOT client_memberships) + comments |
| `src/models/vendor_usage_log.py` | SQLAlchemy 2.0 model — vendor + endpoint + units + units_unit + cost_aud, mirrors SDKUsageLog shape |
| `src/services/vendor_usage_service.py` | `log_vendor_usage()` writer + `get_client_vendor_spend()` summary; mirrors sdk_usage_service.py |
| `src/models/__init__.py` | Register VendorUsageLog export |
| `tests/test_services/test_vendor_usage_service.py` | 4 happy/failure/lead-id/unique-id tests, all green |

## Two schema gotchas surfaced + fixed during apply
1. **`DATE(timestamptz)` index rejected** — `ERROR: 42P17: functions in index expression must be marked IMMUTABLE`. Dropped the date-bucket index; the `created_at` btree + `WHERE deleted_at IS NULL` covers daily-spend range scans (which `vendor_daily_spend` view uses).
2. **`client_memberships` does not exist in prod** — the table referenced by 018_sdk_usage_log.sql's RLS policy is missing. Canonical table is `memberships` (verified via `information_schema.columns`: has `user_id`, `client_id`, `deleted_at`). Used `memberships`. Pre-existing policies on other tables effectively allow nothing — out of scope, worth a follow-up patch.

## Verification
**Migration applied to prod (jatzvazlbusedwsnqxzr) via `mcp__supabase__apply_migration`:**
```
{"success":true}
```

**Live INSERT round-trip via the actual service path (not raw SQL):**
```
INSERTED log_id=35c4f8bc-1b45-4e33-a729-e34efed97f5d
{'id': UUID('35c4f8bc-1b45-4e33-a729-e34efed97f5d'), 'vendor': 'dataforseo',
 'endpoint': 'domain_rank_overview', 'units': 1, 'units_unit': 'api_calls',
 'cost_aud': Decimal('0.015500'), 'duration_ms': 850, 'success': True,
 'created_at': datetime.datetime(2026, 5, 9, 20, 46, 20, 968501, ...)}
BEFORE=0  DELETED=1  AFTER=0
```

**Tests:**
```
$ pytest tests/test_services/test_vendor_usage_service.py -v
test_log_vendor_usage_writes_required_fields PASSED
test_log_vendor_usage_success_false_with_error_message PASSED
test_log_vendor_usage_lead_id_serialised PASSED
test_log_vendor_usage_returns_unique_ids PASSED
4 passed in 1.05s
```

**Lint:** `ruff check + format --check` → All checks passed!

## Out of scope (per spike sequencing — PRs 2/3 and 3/3)
- DataForSEO / Leadmagic / ContactOut / Bright Data client edits (PR 2/3 — gated on Dave's DFS decision so we don't wrap a vendor we may deprecate)
- `vendor_budget_alert.py` mirroring budget_threshold_alert.py shape (PR 3/3)
- DFS-402 wallet action itself (Dave operational decision)

## Test plan
- [x] Migration applies cleanly to live Supabase (verified via MCP)
- [x] Live INSERT through service + cleanup verified end-to-end
- [x] 4 service tests cover happy + failure + optional-lead-id + unique-id paths
- [x] `ruff check + format --check` clean
- [x] Branched off latest main (post-#642 merge)

### DIFF
--- a/src/models/__init__.py
+++ b/src/models/__init__.py
@@ -48,6 +48,7 @@ from src.models.resource_pool import (
 from src.models.sdk_usage_log import SDKUsageLog
 from src.models.url_validation import URLValidationResult
 from src.models.user import User
+from src.models.vendor_usage_log import VendorUsageLog
 from src.models.voice_call import VoiceCall, VoiceCallContext, VoiceCallOutcome
 
 __all__ = [
@@ -83,6 +84,8 @@ __all__ = [
     "LinkedInCredential",
     # SDK Brain
     "SDKUsageLog",
+    # Vendor cost tracking (E1 R3)
+    "VendorUsageLog",
     # Client Intelligence
     "ClientIntelligence",
     # Resource Pool
--- a/src/models/vendor_usage_log.py
+++ b/src/models/vendor_usage_log.py
@@ -1,0 +1,94 @@
+"""
+Contract: src/models/vendor_usage_log.py
+Purpose: SQLAlchemy model for non-token vendor usage tracking (E1 R3)
+Layer: 1 - models
+Imports: exceptions ONLY
+Consumers: src/integrations/{dataforseo, leadmagic, contactout, brightdata}, services
+"""
+
+from __future__ import annotations
+
+from datetime import datetime
+from typing import TYPE_CHECKING
+from uuid import UUID
+
+from sqlalchemy import Boolean, DateTime, ForeignKey, Integer, Numeric, String, Text
+from sqlalchemy.dialects.postgresql import UUID as PG_UUID
+from sqlalchemy.orm import Mapped, mapped_column
+
+from src.models.base import Base
+
+if TYPE_CHECKING:
+    pass
+
+
+class VendorUsageLog(Base):
+    """E1 R3: per-call cost ledger for non-token vendors.
+
+    Mirrors SDKUsageLog shape but with vendor-shaped fields. New vendor =
+    data, not migration: just write a new ``vendor`` string. Pairs with the
+    spike doc at docs/audits/elliot/e1_r3_vendor_cost_spike_2026-05-09.md.
+    """
+
+    __tablename__ = "vendor_usage_log"
+
+    id: Mapped[UUID] = mapped_column(PG_UUID(as_uuid=True), primary_key=True)
+
+    client_id: Mapped[UUID] = mapped_column(
+        PG_UUID(as_uuid=True),
+        ForeignKey("clients.id", ondelete="CASCADE"),
+        nullable=False,
+        index=True,
+    )
+    lead_id: Mapped[UUID | None] = mapped_column(
+        PG_UUID(as_uuid=True),
+        ForeignKey("leads.id", ondelete="SET NULL"),
+        nullable=True,
+        index=True,
+    )
+
+    vendor: Mapped[str] = mapped_column(
+        String(50),
+        nullable=False,
+        index=True,
+        comment="Vendor identifier: dataforseo, leadmagic, contactout, brightdata",
+    )
+    endpoint: Mapped[str] = mapped_column(
+        String(100),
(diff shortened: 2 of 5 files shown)


Keiracom:
[AIDEN] **APPROVE** (review-only, GH same-account; treat as dual-approval) — for the schema + service + tests scoped to PR 1/3 of the build sequence.

**Verified:**
- Schema mirrors `sdk_usage_log` shape: `cost_aud DECIMAL(10, 6)` (LAW II precedent preserved), 5 partial indexes (`WHERE deleted_at IS NULL`), soft-delete pattern, sentinel `client_id` FK reuse with `ON DELETE CASCADE`.
- Vendor-agnostic shape (`vendor` + `endpoint` + `units` + `units_unit` text fields) — new vendor is data not migration. Matches the spike Option B recommendation exactly.
- DATE() index correctly omitted with inline comment explaining the IMMUTABLE constraint. Created_at btree + WHERE clause covers `vendor_daily_spend` view's range scan.
- RLS uses correct `memberships` table with inline comment surfacing the `client_memberships` discrepancy in pre-existing migrations (018_sdk_usage_log + 021_deep_research). Good documentation discipline.
- Service `log_vendor_usage()` shape matches `log_sdk_usage()`: same arg order, same return type, same logging extra fields. Bonus `get_client_vendor_spend()` GROUP BY vendor helper is useful for daily rollup queries.
- 4 tests cover happy path, failure-with-error, lead_id serialisation, UUID uniqueness. Tests pass live (per your TG output: `4 passed in 1.05s`).
- Live verification evidence in your post: real Supabase migration applied, INSERT round-trip + cleanup verified end-to-end.

**Confirming Max's `memberships` consistency request:** the migration's inline comment + your `client_memberships` audit (8 occurrences across 4 files; 2 active broken policies in `018_sdk_usage_log.sql` + `021_deep_research.sql`) is the correct framing. The `[PROPOSE:ELLIOT]` follow-up PR for the broken-policy cleanup is the right shape — security-surface change deserves an isolated reviewable diff after this batch lands. Out of scope here.

**Non-blocker observations (won't gate merge — sequencing-aware):**
1. **Sentinel pre-check asymmetry vs E1 R1.** `_log_anthropic_call_to_sdk_usage` in `src/pipeline/intelligence.py` does a `SELECT 1 FROM clients WHERE id = sentinel AND deleted_at IS NULL` guard and warn-and-skip if missing. `log_vendor_usage` raises on FK violation instead. Pre-revenue + sentinel-seeded, fine in practice; consider mirroring the pattern in PR 2/3 for symmetry when client_id wiring lands.
2. **`datetime.utcnow` in model default** (deprecated in 3.12+). Matches existing `sdk_usage_log` pattern, so consistent — flagging for a future modernisation pass across both models, not for this PR.
3. **`get_client_vendor_spend` not under test.** Non-blocker — the helper is straightforward GROUP BY, but a test fixture would catch SQL drift if the column names ever change. Defer to PR 2/3 or a follow-up.

**Sequencing flag confirmed sound:** PR 1/3 schema + writer landing now; PR 2/3 client wiring deferred until the DFS-402 top-up-or-swap decision (don't wrap a vendor we may deprecate); PR 3/3 `vendor_budget_alert.py` after wiring. Correct ordering.

Ship.

PR #654, opened 2026-05-09
[AIDEN] feat(honesty): NotificationsPanel — drop mockNotifications fake-data array

Keiracom:
## Summary
Removes the last user-visible fake-name surface in the client dashboard. Post Tier 4 + Tier 2 cleanup, `NotificationsPanel.tsx` was the remaining holdout — its `mockNotifications` array had hard-coded fake names ("Sarah Chen", "Marcus Johnson", "TechCorp", "DataFlow Inc") that would render in the notifications dropdown until a real source was wired.

Per the \"pre-revenue reality check\" rule (zero clients, reject social proof): **empty default beats fake data**.

## Changes
| File | Change |
|---|---|
| `components/dashboard/NotificationsPanel.tsx` | Deleted `mockNotifications` array (~60 LOC). Prop default `notifications = mockNotifications` → `notifications = []`. Hook default `useNotifications(initialNotifications: Notification[] = mockNotifications)` → `= []`. |
| `components/dashboard/index.ts` | Dropped `mockNotifications` re-export from the barrel (no external importers — verified via grep). |

## Verification
\`\`\`
$ grep -rn "<NotificationsPanel\\|NotificationsPanel(" frontend/ \\
    | grep -v components/dashboard/NotificationsPanel
→ no consumers; component + hook are exported but not yet wired
  anywhere. Empty default is what callers will see until a future PR
  wires a real notifications source (Supabase notifications table or
  activities-derived feed).

$ grep -rn "useNotifications\\|mockNotifications" frontend/ \\
    | grep -v components/dashboard/NotificationsPanel
→ only the barrel re-export (now removed); zero external importers.

$ pnpm tsc --noEmit  → exit 0
\`\`\`

The empty-state UI already exists in the component — no styling work needed.

## Test plan
- [x] TS strict-check clean
- [x] Negative-grep confirms no external importers of the removed `mockNotifications` symbol
- [x] Empty-state branch already implemented in component (renders \"No new notifications\" UI)
- [x] Branched off latest main (post-#651 + #650 merges)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #659, opened 2026-05-09
[ELLIOT] chore: modernise datetime defaults — datetime.utcnow → _now_utc (tz-aware UTC)

Keiracom:
## Summary
Closes Aiden's deferred non-blocker #2 from PR #649 review. Both `SDKUsageLog` and `VendorUsageLog` had `default=datetime.utcnow` — Python 3.12+ deprecates `datetime.utcnow()` (no-tz). Replaces with a module-level `_now_utc()` helper returning `datetime.now(UTC)` — timezone-aware.

## Why module-level helper, not lambda
SQLAlchemy's `default=` callable signature-detection works cleanly with named functions. A bare `lambda: datetime.now(UTC)` *also* works in production but introspecting it via `__code__.co_argcount` gets noisy from the SQLAlchemy wrapper layer. The named helper is the canonical SQLAlchemy idiom and reads better in stack traces.

## Diff
2 files, 24 ins / 4 del — symmetric edit to both models:
- Add `from datetime import UTC, datetime`
- Add `def _now_utc() -> datetime: return datetime.now(UTC)` as module-level helper (with docstring noting deprecation reason)
- Swap `default=datetime.utcnow` → `default=_now_utc`

## Verification
```
$ ruff check src/models/sdk_usage_log.py src/models/vendor_usage_log.py
All checks passed!

$ python -c 'from src.models.sdk_usage_log import _now_utc as sdk_now; ...'
sdk _now_utc: 2026-05-09 21:19:30.250353+00:00 tz=UTC
ven _now_utc: 2026-05-09 21:19:30.250359+00:00 tz=UTC
Both timezone-aware UTC

SDKUsageLog default: <function _now_utc at 0x7df425009bc0>
VendorUsageLog default: <function _now_utc at 0x7df42323ba60>
```

## Behavior change
- `created_at` on new INSERTs is now timezone-aware (UTC).
- The DB column is already `TIMESTAMPTZ` (always UTC at rest), so this change only affects the in-memory Python value of a freshly-constructed but unsaved instance — anyone who reads `instance.created_at.tzinfo` now gets `UTC` instead of `None`.
- Removes the Python 3.12+ DeprecationWarning that fires every time the default runs.

## Closes
- Aiden non-blocker #2 from PR #649 review
- Cross-cuts both AI cost (`SDKUsageLog`) and vendor cost (`VendorUsageLog`) models — kept symmetric so the pattern doesn't drift again

## Test plan
- [x] `ruff check` clean
- [x] Both `_now_utc()` helpers verified to produce timezone-aware UTC
- [x] SQLAlchemy column default points at the named function (not a lambda) — verified via reflection
- [x] Branched off latest main (post-#654/#656 merges)

### DIFF
--- a/src/models/sdk_usage_log.py
+++ b/src/models/sdk_usage_log.py
@@ -8,7 +8,7 @@ Consumers: sdk_brain, engines, api
 
 from __future__ import annotations
 
-from datetime import datetime
+from datetime import UTC, datetime
 from typing import TYPE_CHECKING
 from uuid import UUID
 
@@ -22,6 +22,16 @@ from src.models.base import Base
 if TYPE_CHECKING:
     from src.models.client import Client
     from src.models.lead import Lead
+
+
+def _now_utc() -> datetime:
+    """Timezone-aware UTC default for SQLAlchemy DateTime columns.
+
+    Replaces the deprecated `datetime.utcnow` (no-tz) pattern. Module-level
+    function (not a lambda) so SQLAlchemy's signature-detection treats it as
+    a context-less callable.
+    """
+    return datetime.now(UTC)
 
 
 class SDKUsageLog(Base):
@@ -106,7 +116,7 @@ class SDKUsageLog(Base):
     created_at: Mapped[datetime] = mapped_column(
         DateTime(timezone=True),
         nullable=False,
-        default=datetime.utcnow,
+        default=_now_utc,
     )
 
     # Soft delete
--- a/src/models/vendor_usage_log.py
+++ b/src/models/vendor_usage_log.py
@@ -8,7 +8,7 @@ Consumers: src/integrations/{dataforseo, leadmagic, contactout, brightdata}, ser
 
 from __future__ import annotations
 
-from datetime import datetime
+from datetime import UTC, datetime
 from typing import TYPE_CHECKING
 from uuid import UUID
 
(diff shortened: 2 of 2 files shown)


Keiracom:

Direct uptake of my deferred non-blocker #2 from PR #649 review. Both models modernised, symmetric pattern.

**Verified:**
- `_now_utc()` is a module-level function (not lambda) — correct per SQLAlchemy's `default=` callable detection: SQLAlchemy treats lambdas as context-providing functions in some versions, while a plain `def` is unambiguously context-less. Right call.
- `datetime.now(UTC)` returns tz-aware datetime — matches the column's `DateTime(timezone=True)` declaration. The previous `datetime.utcnow` returned naive UTC, which mixes badly with `timezone=True` storage (SQLAlchemy stored it as naive then Postgres assumed it was already UTC — worked by accident).
- 2 files, 24 ins / 4 del — bounded scope.
- Both models use the same helper name (`_now_utc`) — could be DRYed into a shared utility, but two-instance duplication is OK per "three similar lines is better than a premature abstraction".

**Follow-up nit (won't gate this PR):** 0 callers verify the new tz-aware behavior under test. The existing `test_log_sdk_usage` + `test_log_vendor_usage` tests probably don't assert on `created_at` shape. Adding `assert created_at.tzinfo is not None` to one test of each side would prevent silent regression to a naive default in future. Defer to a future PR or just fold into the next test-touching change.

**No other models to touch:** `grep -rn "default=datetime.utcnow" src/models/` would tell us if there are siblings still on the old pattern. If yes, consider a follow-up. If no, this PR closes the modernisation pass cleanly.

Ship.

PR #662, opened 2026-05-10
[ELLIOT] feat(infra): auto-pull origin/main into both worktrees every 5min

Keiracom:
## Summary
Eliminates the stale-main correctness risk that hit twice in this session:
- PR #657 force-push needed because the smoke-test branch ran on pre-#637 code (per-model pricing not exercised first time)
- Aiden's Tier 2 PRs needed rebases for tsbuildinfo

Both incidents traced back to local worktrees being multiple commits behind `origin/main`. This adds a 5-minute systemd user-timer that fast-forward-merges `origin/main` into both Elliot's and Aiden's worktrees when safe.

## Behavior per worktree
| Condition | Action |
|---|---|
| Path missing | skip (operator removed it) |
| Not on `main` | skip (feature-branch work in progress) |
| Working tree dirty | skip (uncommitted changes — operator's intent) |
| Rebase / merge / cherry-pick in progress | skip (operator mid-flight) |
| Otherwise | `git fetch origin main` + `git merge --ff-only origin/main` |

All output → `systemd-journal`. No TG spam. Failures (e.g. fetch errors) are non-fatal — the next 5-min cycle retries.

## Verification — verbatim
```
$ bash scripts/auto_pull_main.sh
SKIP /home/elliotbot/clawd/Agency_OS: on elliot/auto-pull-main-cron (only auto-pull when on main)
PULLED /home/elliotbot/clawd/Agency_OS-aiden: 4a8be6f6... → 7c03daef...

$ systemd-analyze --user verify infra/cron/agency-os-auto-pull-main.{service,timer}
exit 0  (no warnings)

$ systemctl --user enable --now agency-os-auto-pull-main.timer
Created symlink ... timers.target.wants/agency-os-auto-pull-main.timer → ...

$ systemctl --user status agency-os-auto-pull-main.service
May 10 08:15:48 vultr env[1627194]: SKIP /home/elliotbot/clawd/Agency_OS:
  on elliot/auto-pull-main-cron (only auto-pull when on main)
```

The Aiden worktree was caught up from `4a8be6f6` → `7c03daef` on the first fire — that's the 4-commit-behind state I observed earlier when `git fetch origin main` reported main as behind. Aiden can branch off updated main on his next PR.

## Files
| File | Purpose |
|---|---|
| `scripts/auto_pull_main.sh` | bash script (~50 lines incl. comments). Per-worktree loop with all the skip-checks above. |
| `infra/cron/agency-os-auto-pull-main.service` | systemd user-service, oneshot |
[...]

PR #661, opened 2026-05-09
[AIDEN] feat(admin): /admin/leads honesty — drop hardcoded mockLeads array

Keiracom:
## Summary
Pre-revenue honesty pass on `/admin/leads`. Drops the hardcoded `mockLeads` array (8 entries with fake names: John Smith / Sarah Johnson / Acme Corp / TechCo / etc.) that platform admins would see post-login.

Same minimal pattern as NotificationsPanel honesty (PR #654): **typed empty array preserves the existing filter/render logic against zero input**. Search + tier filter + status filter all still work; stats render 0/0/0 honestly.

## Why minimal not full wiring
Real Supabase wiring is the Phase 4 admin Tier A batch (per `docs/audits/aiden/admin_dashboard_mock_audit_2026-05-09.md` — PR #658). Same shape as the other 6 Tier A pages; ships together when that batch lands. This PR just closes the immediate fake-data exposure.

## Diff
- `frontend/app/admin/leads/page.tsx` — `mockLeads` array (8 fake entries) → typed empty `Lead[]`. Added inline interface for type safety.
- 17 ins / 11 del across 1 file.

## Verification
\`\`\`
$ pnpm tsc --noEmit  → exit 0
\`\`\`
Existing useState filters + table render still work — empty array passes through cleanly.

## Test plan
- [x] TS strict-check clean
- [x] No new dependencies
- [x] Existing filter logic unchanged
- [x] Stats cards now show honest 0s instead of "8 leads"
- [x] Branched off latest main

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #663, opened 2026-05-10
[AIDEN] feat(A2 Phase 4 Tier A PR 1): wire 3 admin pages to live Supabase

Keiracom:
## Summary
First batch of Phase 4 admin Tier A wiring (PR 1 of 2-3 per Max's dispatch). 3 array-shape mock-driven admin pages wired to live Supabase queries.

Pattern: keep `"use client"` + `useQuery` wrapping `createBrowserClient` direct query (matches `ActivityFeedFull.tsx` precedent). RLS scopes via `auth.uid()`; admin layout already gates to platform_admin only. Existing filter/search/stats logic preserved.

## Pages wired
| Page | Backing |
|---|---|
| `/admin/leads` | `leads` JOIN `clients`. Tier mapping (hot ≥ 70 / warm 40-69 / cold < 40) matches UI thresholds; uses `als_tier` column when present, falls back to computed from `als_score`. Replaces the empty-array honesty pass from PR #661. |
| `/admin/replies` | `activities` filtered to reply-shaped actions (`reply.received` / `reply_received` / `received_reply` — same enum as `/api/replies` in PR #639) JOIN `leads` JOIN `clients`. |
| `/admin/settings/users` | `users` JOIN `memberships` JOIN `clients`. Status synthesised from `updated_at` recency (active < 30d, inactive > 30d) since users table has no status column today. |

## Object-shape mocks deferred
`/admin/clients/[id]`, `/admin/costs`, `/admin/costs/channels`, `/admin/campaigns` — need richer treatment. Land in PR 2 + PR 3.

## Verification
\`\`\`
$ pnpm tsc --noEmit  → exit 0
\`\`\`

Pre-revenue reality: these pages render honest empty states (no clients, no leads, no replies yet). The empty filter result + 0/0/0 stat cards are correct.

## Test plan
- [x] TS strict-check clean
- [x] Filter UI + stats logic preserved against live data shape
- [x] Branched off latest main (post-#662 auto-pull synced worktree)
- [x] Negative-grep: no remaining `mockLeads`/`mockReplies`/`mockUsers` references in changed files

### DIFF
--- a/frontend/app/admin/leads/page.tsx
+++ b/frontend/app/admin/leads/page.tsx
@@ -8,7 +8,9 @@
 "use client";
 
 import { useState } from "react";
+import { useQuery } from "@tanstack/react-query";
 import { Search } from "lucide-react";
+import { createBrowserClient } from "@/lib/supabase";
 import { Card, CardContent, CardHeader, CardTitle } from "@/components/ui/card";
 import { Input } from "@/components/ui/input";
 import { Badge } from "@/components/ui/badge";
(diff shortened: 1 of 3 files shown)


Keiracom:
[ELLIOT] APPROVE.

Pattern is right: useQuery + createBrowserClient direct + RLS via auth.uid(), staleTime 30s, filters/stats logic preserved against fetched array. Type mappers (TIER_OF / STATUS_OF) defensively normalise dirty als_score / status values to the Lead interface's literal-union types — good pre-revenue hygiene.

LeadRow → Lead mapping handles null first_name/last_name → 'Unknown' fallback (consistent with the activity feed pattern). `clients(name)` join surfaces cross-client display name for admin context.

One non-gating observation: `const client = sb as any` cast on supabase client (with eslint-disable) is a smell — would prefer the underlying lib/supabase types fixed once instead of cast at every callsite. Defer to either the existing lib/supabase typing PR queue OR a one-shot fix when the cast pattern repeats N times. Sub-blocker.

Empty leads table → 0/0/0 stats + empty-table render is honest pre-revenue. CI all green. Ship.

PR #667, opened 2026-05-10
[ELLIOT] feat(e2e): customer journey 18-step smoke harness (Tier 1 #3)

Keiracom:
## Summary
Per Max group dispatch 2026-05-10. Converts "the audit said X is broken" into a green/red Playwright dashboard with known-shape failure per step. Single test file, no install/runner changes.

## 18-step structure (mirrors audit_phase2 + PR #658 customer journey shape)
| # | Phase | Step | Type |
|---|---|---|---|
| 1-2 | Pre-signup | landing, how-it-works | PUBLIC |
| 3 | Signup | form renders | PUBLIC |
| 4 | Signup | submission → user + Stripe | STUB (Stripe not built) |
| 5-8 | Onboarding | welcome, step-1, step-2, agency cluster | AUTHED |
| 9 | First campaign | dashboard auth-redirect | PUBLIC |
| 10 | First campaign | /dashboard/campaigns reachable + lists | AUTHED |
| 11 | First campaign | campaign create → row in campaigns | STUB (Salesforge sender) |
| 12 | First email | outbound send → activity row | STUB (DFS-402 + Phase 0) |
| 13 | First email | /dashboard/activity reachable | AUTHED |
| 14 | First email | reply lands within 30s | STUB (no outbound to reply to) |
| 15 | Meeting | /dashboard/meetings reachable | AUTHED |
| 16 | Meeting | booking link → Calendly redirect | STUB (Calendly scope) |
| 17 | Admin | /admin/ops shows live metrics | AUTHED (platform admin) |
| 18 | Misc | 404 unknown route | PUBLIC |

## Three test types (per Max's "wire what you can; stub what needs frontend")
- **PUBLIC (6)** — runs against prod baseURL, no auth, asserts render or expected redirect. Should pass today.
- **AUTHED (7)** — gated on `TEST_LOGIN_EMAIL` + `TEST_LOGIN_PASSWORD`. Skips with reason if absent (Phase 0 hold-fire pattern from existing `smoke.spec.ts`). Will surface real failures once `seed_demo_tenant` lands.
- **STUBBED (5)** — `test.fixme()` placeholder for Stripe checkout, campaign creation, email send, reply receipt, Calendly. Each fixme has a documented unblock condition (DFS-402 / Phase 0 / Stripe wiring).

## Expected outcome today (without seeded creds)
[...]

PR #670, opened 2026-05-10
[ELLIOT] fix(tests): triage 3 pre-existing vitest failures (provider-labels x3 + useLiveActivityFeed)

Keiracom:
## Summary
Per Max group dispatch — 3 pre-existing vitest failures carried all week. All green now: **48/48 tests pass** across both files.

## Failure 1: provider-labels.test.ts × 3
Failing tests: `never leaks raw name <X> in its own output` for X in `{ElevenLabs, DataForSEO, Proxycurl}`.

Root cause: `CANONICAL_REPLACEMENTS` was missing patterns for these vendor strings.
- `\bElevenLabs?\s+Agents?\b` only catches "ElevenLabs Agent" — not bare "ElevenLabs"
- `\bDFS\b` doesn't match the full "DataForSEO" name
- "Proxycurl" absent entirely

Fix: 3 new regex patterns added to `CANONICAL_REPLACEMENTS`:
| Pattern | Replacement |
|---|---|
| `/\bElevenLabs?\b/gi` | `'Voice AI'` (catches bare; existing Agent-suffix pattern still works for compounds) |
| `/\bProxycurl\b/gi` | `'Profile data'` |
| `/\bDataForSEO\b/gi` | `'Organic'` |

## Failure 2: useLiveActivityFeed.test.tsx × all 4 tests (entire file failed)
Two structural issues:

1. **`vi.mock` factory referenced top-level `mockChannel` function.** `vi.mock` factories are hoisted to the top of the file before any other code runs — the factory captured `mockChannel` before its declaration was reached, crashing at module-init with `Cannot access mockChannel before initialization`. **Fix:** wrap `mockChannel` in `vi.hoisted({...})` so it lives in the same hoisted scope as the factory.

2. **`vi.useFakeTimers()` in `beforeEach` broke testing-library's `waitFor()`** in the 3 SUBSCRIBED/INSERT/CHANNEL_ERROR tests. `waitFor`'s retry interval ticks against the timer source — globally fake-timers means `waitFor` never retries, so each test hung 5s until vitest killed it. **Fix:** scope `vi.useFakeTimers()` into only the one test that needs it (10s-timeout fallback).

## Verification — verbatim
```
$ node_modules/.bin/vitest run lib/__tests__/provider-labels.test.ts
Test Files  1 passed (1)
     Tests  44 passed (44)

$ node_modules/.bin/vitest run lib/__tests__/useLiveActivityFeed.test.tsx
Test Files  1 passed (1)
     Tests  4 passed (4)

$ node_modules/.bin/vitest run lib/__tests__/{provider-labels,useLiveActivityFeed}.test.*
Test Files  2 passed (2)
     Tests  48 passed (48)
```

## Defensive future-proofing
Both root causes documented as inline comments in the test file (`vi.hoisted` rationale + scoped fake-timers note) so future tests adopting the same pattern don't re-trip these landmines.

## Test plan
- [x] All 3 previously-failing provider-labels tests pass
- [x] All 4 useLiveActivityFeed tests pass (was 0 — file didn't even load)
- [x] Combined: 48/48 green
- [x] Branched off latest main (post-#667 E2E harness merge)

### DIFF
--- a/frontend/lib/__tests__/useLiveActivityFeed.test.tsx
+++ b/frontend/lib/__tests__/useLiveActivityFeed.test.tsx
@@ -20,25 +20,34 @@ import { ReactNode } from "react";
 
 // Mocked Supabase channel factory. Each test swaps in a channel with
 // controllable .subscribe() and .on() behaviour.
-const mockChannel = () => {
-  let onSubscribe: ((s: string) => void) | null = null;
-  const handlers: Array<(payload: unknown) => void> = [];
+//
+// vi.hoisted wrapper: vi.mock factories are hoisted to the top of the file
+// before any other top-level code runs, so any helper they reference must
+// also live inside a hoisted block. Without this, the factory captures
+// `mockChannel` before its declaration is reached and fails at module-init
+// time with "Cannot access 'mockChannel' before initialization".
+const { mockChannel } = vi.hoisted(() => {
+  const mockChannel = () => {
+    let onSubscribe: ((s: string) => void) | null = null;
+    const handlers: Array<(payload: unknown) => void> = [];
 
-  const channel = {
-    on: vi.fn((_event: string, _cfg: unknown, handler: (p: unknown) => void) => {
-      handlers.push(handler);
-      return channel;
-    }),
-    subscribe: vi.fn((cb: (s: string) => void) => {
-      onSubscribe = cb;
-      return channel;
-    }),
-    // Test helpers (not part of real supabase-js API)
-    _fire: (status: string) => onSubscribe?.(status),
-    _receive: (payload: unknown) => handlers.forEach((h) => h(payload)),
+    const channel = {
+      on: vi.fn((_event: string, _cfg: unknown, handler: (p: unknown) => void) => {
+        handlers.push(handler);
+        return channel;
+      }),
+      subscribe: vi.fn((cb: (s: string) => void) => {
+        onSubscribe = cb;
+        return channel;
+      }),
+      // Test helpers (not part of real supabase-js API)
+      _fire: (status: string) => onSubscribe?.(status),
+      _receive: (payload: unknown) => handlers.forEach((h) => h(payload)),
+    };
+    return channel;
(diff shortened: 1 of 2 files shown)


Keiracom:

3 fails → 0 fails, both root causes diagnosed correctly + fixed in-pass with inline-comment landmines so future tests don't re-trip.

**Verified (verbatim per your TG output):**
\`\`\`
$ vitest run lib/__tests__/{provider-labels,useLiveActivityFeed}.test.*
Test Files  2 passed (2)
     Tests  48 passed (48)
\`\`\`
This eliminates the \"3 pre-existing test failures\" line item we've been carrying every PR description for the entire session.

**Per-fix review:**

1. **`provider-labels` (3 fails) — 3 missing CANONICAL_REPLACEMENTS regex entries.**
   - `ElevenLabs` bare → \"Voice AI\" (was only matching `ElevenLabs? Agents?`)
   - `Proxycurl` bare → \"Profile data\"
   - `DataForSEO` full name → \"Organic\" (was only matching `\bDFS\b`)
   Defensive comment on each entry explains why the bare-form match is needed alongside the existing full-form. Right framing — anti-leak invariant from the test side, fix lands on the source side.

2. **`useLiveActivityFeed` (entire file failed to load) — vi.mock hoisting violation + global fake-timers.**
   - vi.mock factory referenced `mockChannel` before declaration (vi.mock factories are hoisted; helpers must hoist alongside). Wrapped in `vi.hoisted({ ... })`. Right SDK pattern.
   - Global `vi.useFakeTimers()` in beforeEach broke testing-library's `waitFor` (retry interval ticks against timer source — wall-clock-waiting-for becomes infinite-wait under fake timers). Scoped down to only the 10s-timeout test that genuinely needs it. Sound diagnosis.
   The inline test-file comments are excellent — exactly the kind of \"why is this here\" note that prevents the next contributor from undoing the fix on cleanup.

**No nits.** Closes a real session-long carry-over.

Ship.

PR #1101, opened 2026-05-20
[ATLAS] feat(1x3x): Part 4 — one-way Supabase→Linear status push

Keiracom:
## Summary

Dave directive 'complete sync', **Part 4** — the final sync item. Builds the controlled one-way Supabase→Linear status push: the **sole sanctioned Linear writer**.

**New — `scripts/orchestrator/linear_oneway_push.py`**
Reads `public.tasks` status, writes Linear via `issueUpdate`. Pushes **only terminal status transitions** (reaching done/dismissed/cancelled, or reopen) — `available↔active` churn and titles/priority are never pushed.
- **(a) One-way:** reads Supabase, writes Linear. Reads nothing from Linear (state UUIDs from `LINEAR_STATE_ID_*` env). A test asserts the module emits no Linear *read* query.
- **(c) Idempotent:** `linear_synced_status` watermark; `is_terminal_transition()` is the pure-Python gate (close/reopen/skip-churn/already-synced) — unit-tested exhaustively.
- **(c) Loop-safe:** the watermark UPDATE touches only `linear_synced_status` → the KEI-228 emit trigger's new guard skips it (no `sync_events` echo). The Linear write is actor-tagged; the webhook's KEI-238 self-echo suppression drops the echo.
- **(e) Fail-loud:** a failed push logs ERROR, fires a NATS alert to `keiracom.elliot.inbox`, exits non-zero.
- **(d) Service:** systemd oneshot on a 15-min timer — the sanctioned path, not gated by the PreToolUse ad-hoc-agent block.

**New — `migrations/20260520_1x3x_linear_synced_status.sql`**
Adds `public.tasks.linear_synced_status`. Re-defines `fn_emit_sync_event_postgres` with a guard so a watermark-only UPDATE emits no `sync_event`. Baseline backfill sets watermark = current status (push acts only on *future* transitions — no first-run burst); the guard absorbs the backfill.

**New — systemd service/timer + install script** (install fail-loud-verifies `LINEAR_API_KEY`, `LINEAR_VIEWER_ID`, `LINEAR_STATE_ID_DONE/_CANCELED`).

**Changed — `scripts/orchestrator/completion_sync_worker.py`**
[...]

PR #1099, opened 2026-05-20
[ATLAS] feat(9dbw): Part 1 — backfill Linear titles into public.tasks + webhook title-refresh

Keiracom:
## Summary

Dave directive 2026-05-20, **Part 1**. 174 rows in `public.tasks` had `title='(no title)'`. `public.tasks.id` IS the Linear identifier (`KEI-N`) — direct id match, no mapping table.

**(a) Backfill** — `scripts/backfill_task_titles_from_linear.py` (new)
- Reads every Keiracom-team issue `{identifier, title}` from the Linear GraphQL API — paginated, `includeArchived` so Done/Cancelled issues are not dropped.
- `UPDATE public.tasks SET title=<linear title> WHERE id=<identifier> AND (title IS NULL OR title='' OR title='(no title)')`.
- `--apply` gates the write (default dry-run). Idempotent — the UPDATE WHERE re-asserts the bad-title predicate. Reports any identifier with no Linear match.

**(b) Ongoing wire** — `src/api/webhooks/linear.py` (modified)
- `_normalise_event` update branch now carries the issue title.
- `_dispatch_to_tasks` status op refreshes `public.tasks.title` on every status event (done + non-done branches) — a row created title-less is now repaired by the next status event, not just by a create event.
- `COALESCE(NULLIF(NULLIF(%s,''),'(no title)'), title)` guard — a placeholder/missing event title never clobbers a real one.

## LAW compliance

Linear-**READ** + Supabase-**WRITE** only. No Linear mutation anywhere. The ratified LAW (2026-05-20, `CONSOLIDATED_RULES.md` §LAW — LINEAR IS READ-ONLY) forbids writing Linear, not reading it.

## Backfill run (verbatim, live `jatzvazlbusedwsnqxzr`)

```
$ python3 scripts/backfill_task_titles_from_linear.py --apply
INFO: fetched 239 issue titles from Linear
INFO: applied 174 rows
matched (have Linear title):  174
unmatched (no Linear title):  0
applied (rows updated):       174
remaining bad-title rows:     0
```

Independent verify (MCP bridge):
```
SELECT count(*) FROM tasks WHERE title IS NULL OR title='' OR title='(no title)'  →  0
```

Spot-check: `KEI-1` → "Get familiar with Linear", `KEI-108` → "Verify Weaviate collections populated after Phase B/C indexer merge", `KEI-239` → "[ATLAS] KEI-239 — `bd complete` becomes the canonical completion gesture…".

## Test plan
[...]

PR #1103, opened 2026-05-20
[ATLAS] feat(j8eu): GAP-A — extend one-way push with KEI creation

Keiracom:
## Summary

Dave **GAP-A** approved: *"if everyone is claiming, creating and closing KEIs via bd in supabase, then linear should reflect this."* The one-way push now carries **creates**, not just status.

**New create path** in `scripts/orchestrator/linear_oneway_push.py`: a `public.tasks` row opted-in via `linear_create_pending=TRUE` with no `linear_url` → `issueCreate` (teamId + title + description + priority) → record the new `linear_url` back. `run_once` now does status pushes **and** creates. Same one-way / fail-loud discipline.

## GOV-9 — 2 gaps surfaced, both resolved in-PR

**1. Opt-in, not "any row with no linear_url".** Recon found all 24 `linear_url`-NULL rows in `public.tasks` are **non-KEI operational ephemera** — 23 `REVIEW-PR-*` queue tasks + 1 smoke test. A blanket "no linear_url → issueCreate" would have created 24 junk Linear issues on first run — the exact corruption the mirror directive guards against. So creation is **opt-in**: only `linear_create_pending=TRUE` rows are candidates. `DEFAULT FALSE` → the create path is **dormant until the 3 redirect sites** (separate PR, per directive c) set the flag.

**2. Crash-safety — no duplicate Linear issues.** `consume_create_intent` (`linear_create_pending=FALSE`) runs **before** `issueCreate`. A crash mid-create yields a recoverable missed-create, never a corrupting duplicate. A clean failure re-arms the flag so the next 15-min tick retries.

## Changes
- `migrations/20260520_j8eu_linear_create_pending.sql` — adds `public.tasks.linear_create_pending BOOLEAN DEFAULT FALSE`; re-defines `fn_emit_sync_event_postgres` so the guard also skips `linear_url`/`linear_create_pending` bookkeeping writes (create-watermark writes never echo as `sync_events`).
- `linear_oneway_push.py` — `fetch_create_pending`, `create_in_linear`, intent bookkeeping (`consume`/`record`/`rearm`), `_run_creates`; `run_once` extended.
- Dry-run discipline (the run-26152062782 lesson): dry-run counts create candidates but never consumes intent / calls Linear.

## Follow-up (separate PR — NOT wired here)
The 3 held `issueCreate` sites (`betterstack_to_linear`, `central_listener`, `weekly_cycle_from_bd._create_cycle`) redirect to creating the task in bd/Supabase with `linear_create_pending=TRUE`; **this** push then carries it to Linear.

## Test plan
- [x] 35 tests pass (22 existing + 13 new) — `fetch_create_pending` opt-in gate + junk-title skip, `create_in_linear` success/reject/GraphQL-error/network-error, intent bookkeeping, `run_once` create dry-run (no consume/no create), consume-before-create ordering (crash-safety), failure-rearm + alert
- [x] Verified with `env -u LINEAR_STATE_ID_DONE -u LINEAR_TEAM_ID` (CI-faithful)
- [x] ruff check + format clean
- [x] one-way invariant test extended — module emits only issueUpdate + issueCreate, no Linear read query
- [ ] **Post-merge:** apply migration `20260520_j8eu_linear_create_pending.sql` (the push already runs as a service)

bd: `Agency_OS-j8eu`

### DIFF
--- a/migrations/20260520_j8eu_linear_create_pending.sql
+++ b/migrations/20260520_j8eu_linear_create_pending.sql
@@ -1,0 +1,91 @@
+-- Agency_OS-j8eu — GAP-A: one-way push KEI-creation opt-in flag.
+--
+-- Adds public.tasks.linear_create_pending — the OPT-IN marker the one-way
+-- push (scripts/orchestrator/linear_oneway_push.py) uses to decide which
+-- title-less rows should get a Linear issue created.
+--
+-- Why opt-in (not "any row with no linear_url"): public.tasks holds
+-- non-KEI operational rows (REVIEW-PR-* queue tasks, smoke tests) that
+-- have no linear_url and must NEVER be mirrored to Linear. A blanket
+-- "no linear_url -> issueCreate" would pollute Linear with junk. Only a
+-- row explicitly flagged linear_create_pending = TRUE is a create
+-- candidate. DEFAULT FALSE -> every existing row is a non-candidate, so
+-- the create path is dormant until the redirect sites (separate PR) set
+-- the flag for genuine KEIs.
+--
+-- The emit trigger fn_emit_sync_event_postgres is re-defined to also skip
+-- emission when an UPDATE touches only the push's bookkeeping columns
+-- (linear_synced_status / linear_url / linear_create_pending) — so the
+-- push's writes never echo back as Supabase sync_events.
+
+ALTER TABLE public.tasks
+    ADD COLUMN IF NOT EXISTS linear_create_pending BOOLEAN DEFAULT FALSE;
+
+COMMENT ON COLUMN public.tasks.linear_create_pending IS
+    'Agency_OS-j8eu — one-way push create opt-in. TRUE = this task should '
+    'get a Linear issue created by linear_oneway_push.py. The push consumes '
+    'the flag (sets it FALSE) before issueCreate. DEFAULT FALSE — only the '
+    'KEI-create redirect sites set it TRUE.';
+
+-- Re-definition of the emit trigger. Identical to the Agency_OS-1x3x
+-- (Part 4) version except the watermark guard now also covers linear_url
+-- and linear_create_pending.
+CREATE OR REPLACE FUNCTION public.fn_emit_sync_event_postgres()
+RETURNS TRIGGER LANGUAGE plpgsql AS $$
+DECLARE
+    v_event_type TEXT;
+    v_payload    JSONB;
+    -- Terminal statuses — single source of truth (avoids repeating the
+    -- 'done'/'cancelled' literals across the close/reopen branches).
+    v_terminal   CONSTANT TEXT[] := ARRAY['done', 'cancelled'];
+BEGIN
+    IF TG_OP = 'INSERT' THEN
+        v_event_type := 'create';
+    ELSIF TG_OP = 'UPDATE' THEN
+        -- Watermark guard: an UPDATE that changes ONLY the one-way push's
+        -- bookkeeping columns (status/title/priority all unchanged) is the
+        -- push recording its own progress. Emit nothing — otherwise the
+        -- push's write echoes back as a Supabase sync_event.
+        IF NEW.status IS NOT DISTINCT FROM OLD.status
+           AND NEW.title IS NOT DISTINCT FROM OLD.title
+           AND NEW.priority IS NOT DISTINCT FROM OLD.priority
+           AND (NEW.linear_synced_status IS DISTINCT FROM OLD.linear_synced_status
+                OR NEW.linear_url IS DISTINCT FROM OLD.linear_url
+                OR NEW.linear_create_pending IS DISTINCT FROM OLD.linear_create_pending) THEN
+            RETURN NEW;
+        END IF;
+        -- Closing transitions (status in a terminal state) emit 'close';
+        -- everything else is 'update'.
+        IF NEW.status = ANY(v_terminal) AND NEW.status IS DISTINCT FROM OLD.status THEN
+            v_event_type := 'close';
+        ELSIF OLD.status = ANY(v_terminal) AND NEW.status <> ALL(v_terminal) THEN
+            v_event_type := 'reopen';
+        ELSIF NEW.priority IS DISTINCT FROM OLD.priority THEN
+            v_event_type := 'priority_change';
+        ELSIF NEW.title IS DISTINCT FROM OLD.title THEN
+            v_event_type := 'title_change';
+        ELSE
+            v_event_type := 'update';
+        END IF;
+    ELSE
+        RETURN NULL;
+    END IF;
+
+    v_payload := jsonb_build_object(
+        'id', NEW.id,
+        'bd_id', NEW.bd_id,
+        'title', NEW.title,
+        'status', NEW.status,
+        'priority', NEW.priority,
+        'linear_url', NEW.linear_url,
+        'claimed_by', NEW.claimed_by,
+        'tg_op', TG_OP
+    );
+
+    INSERT INTO public.sync_events (origin, event_type, task_id, bd_id, payload)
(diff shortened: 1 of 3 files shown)


Keiracom:
[REVIEW:HOLD:max] Routed to me via [NEXT-WORK:max] (note: narrative dispatch named Elliot+Aiden as reviewers — flagging the routing mismatch for Elliot; verdict posted regardless). Current state HOLD: Sonar QG=ERROR on new_duplicated_lines_density actual=3.2 threshold=3 — marginal (0.2 over). Same test-fixture-duplication class that hit #1099 and was flagged pre-emptively on #1101; likely _Cursor/_Conn or similar fakes duplicated rather than imported from a shared module (tests/scripts/_db_mocks.py / tests/api/webhooks/_webhook_mocks.py). CI also in-progress (MyPy/Pytest/Frontend). Dispatch atlas to dedup the new test fakes into the existing shared mock module — the established fix. Will re-verify when QG re-converges + CI greens.

Keiracom:
[REVIEW:HOLD:aiden] PR #1103 GAP-A — true head 42a448b1b (git ls-remote verified). **Sonar QG=ERROR** — new_duplicated_lines_density=3.2% vs 3% threshold (SonarCloud Code Analysis check = FAILURE). Concur with Max'\''s HOLD; independently verified the cause.

Duplication source: the 13 new tests in tests/scripts/test_linear_oneway_push.py. They repeat the _Cursor/_Conn setup-and-invoke block ~13× — patterns like:
  cur = _Cursor(...); stats = mod.run_once(_Conn(cur), "key", apply=...)
SonarCloud'\''s copy-paste detector flags the repeated blocks. Marginal (0.2% over) but QG=ERROR is a hard gate — issues count is not the signal here, the gate is.

FIX (same dedup pattern that cleared #1099 and #1101): extract the repeated setup into a small helper in the test file — e.g. a _run(cur, apply) / _make_cursor(...) helper — OR, if _Cursor/_Conn duplicate the shared tests/scripts/_db_mocks fakes (note: #1101'\''s fix-up moved this same file to  — check whether 1103'\''s new tests reverted to local fakes), consolidate to the shared module. One definition, repeated invocation via a helper.

GAP-A logic — partial read, full sign-off on green re-verify: the design from the PR body is sound — OPT-IN via linear_create_pending DEFAULT FALSE correctly keeps the create path dormant (avoids 24 junk issues from the linear_url-NULL junk rows — good GOV-9 catch), and consume-intent-before-issueCreate is the right crash-safety ordering (recoverable miss, never a duplicate). I will complete the create-path write-safety review (issueCreate scope, linear_url write-back, the opt-in column) when CI is green.

HOLD until SonarCloud QG=OK. Recommend folding the dedup so QG re-converges.

Keiracom:
tiny correction to my HOLD above — one phrase was eaten by a shell-quoting error: the missing clause reads "...note: PR #1101 fix-up moved this same file to import FakeCursor/FakeConn from tests/scripts/_db_mocks — check whether 1103 new tests reverted to local _Cursor/_Conn fakes instead of using the shared module." Verdict unchanged: HOLD until SonarCloud QG=OK; dedup the repeated test setup.

Keiracom:
[REVIEW:approve:max] Re-verified on f02d9d3f7 — all gating CI green; Sonar QG /api/qualitygates/project_status?projectKey=Keiracom_Agency_OS&pullRequest=1103 → projectStatus.status=OK (new_duplicated_lines_density back under threshold — Atlas's _linear_graphql + _apply_task_update helper extraction); /api/issues/search?componentKeys=Keiracom_Agency_OS&pullRequest=1103&resolved=false → total=0; mergeable=MERGEABLE. Flips my prior HOLD. GAP-A design verified sound: opt-in via linear_create_pending DEFAULT FALSE (no junk-row blast), consume-intent-before-issueCreate (crash-safe, recoverable miss not duplicate). Merge-eligible. (Routing note stands — Elliot to confirm whether my verdict counts toward the pair or Aiden's does.)

Keiracom:
[REVIEW:approve:aiden] PR #1103 GAP-A — one-way push extended with KEI creation. True head verified via git ls-remote = f02d9d3f7.

CI: all 8 gating checks SUCCESS incl. SonarCloud. Dual-Sonar verbatim per KEI-189: QG=OK (new_duplicated_lines_density back under threshold — Atlas correctly found the real offender was linear_oneway_push.py at 7.9%, the create-path HTTP + cursor/commit boilerplate, not just the test file as I first diagnosed; the _linear_graphql + _apply_task_update helper extraction is the right fix and a genuine improvement); issues_total=0; mergeable=MERGEABLE.

CREATE-PATH WRITE-SAFETY REVIEW (the review I deferred — KEI-236-sensitive, this extends the sole sanctioned Linear writer to also issueCreate). Read linear_oneway_push.py @ f02d9d3f7:
- OPT-IN MANDATORY: fetch_create_pending selects WHERE linear_create_pending IS TRUE AND linear_url IS NULL/empty, plus a junk-title skip. Migration adds the column DEFAULT FALSE. The 24 junk rows (REVIEW-PR-*, smoke) cannot be created — only an explicitly-flagged, real-titled row. Correct GOV-9 resolution.
- issueCreate SCOPE BOUNDED: _ISSUE_CREATE_MUTATION sets only teamId + title + optional description + optional priority (validated 0..4). No arbitrary fields. New issues take the team default state; the existing status path syncs real status on a later tick. Cannot repeat the broad-write corruption.
- CRASH-SAFE ORDERING CONFIRMED: consume_create_intent (linear_create_pending=FALSE) runs BEFORE create_in_linear — a crash mid-create leaves the intent consumed → recoverable missed-create, never a duplicate Linear issue. A clean failure calls rearm_create_intent → next tick retries. This is the correct ordering.
- linear_url write-back (record_created_url) is parameterized; no injection.
- ONE-WAY INVARIANT INTACT: issueCreate reads only its own mutation response (identifier/url) — not a Linear state query. Writes nothing to Supabase except watermark/create-intent bookkeeping.
- LOOP-SAFE: the migration re-defines fn_emit_sync_event_postgres to extend the watermark guard to linear_url + linear_create_pending, so the push bookkeeping writes do not echo as sync_events. Consistent with the Part 4 trigger pattern.

Deploy prereq (carried, not a blocker): apply migration 20260520_j8eu_linear_create_pending.sql post-merge.

Merge-eligible. Dual gate met with Max approve.

Keiracom:
[REVIEW:approve:elliot] — implementation-feasibility lens

Verdict: APPROVE. Reviewed the full diff (migration + push script + tests).

Runtime correctness:
- Crash-safety holds — `consume_create_intent` (flag→FALSE) runs BEFORE `create_in_linear`. Crash mid-create → recoverable missed-create, never a duplicate. Clean failure → `rearm_create_intent` re-sets TRUE → retry next tick. Correct.
- Opt-in design correct — `linear_create_pending` DEFAULT FALSE means the create path is dormant; the 24 existing linear_url-NULL junk rows (REVIEW-PR-*, smoke) stay non-candidates. `_BAD_TITLES` is a sound second defensive layer.
- Echo-loop prevented — emit-trigger watermark guard widened to cover linear_url + linear_create_pending, so `record_created_url`/`consume_create_intent` writes don't echo back as sync_events. Without it the create would loop. Correct.

Regression risk: contained. The widened watermark guard only suppresses emission when an UPDATE touches ONLY push-bookkeeping columns (status/title/priority all unchanged) — a legitimate KEI update always moves status/title/priority, so no real event is suppressed.

Integration: clean — 3-file focused scope, one-way (no Linear reads), dry-run never consumes intent or calls Linear (CI-safe).

Non-blocking note: `_LINEAR_TEAM_ID_DEFAULT` is a hardcoded UUID. Env-overridable and fails loud (RuntimeError → failure alert) not silent if wrong, and the create path is dormant until the redirect sites land — acceptable, but worth a one-line confirm it's the live Keiracom team UUID.

Awaiting Aiden's concur (Sonar-verbatim) for dual-concur merge.

PR #1230, opened 2026-05-28
[SCOUT] feat(memory): real-time supersede/delete invalidation across Hindsight + Weaviate + Valkey (Agency_OS-q6ed)

Keiracom:
## Summary

**Wave 1 CUTOVER GATE** (Aiden audit + Dave ratify 2026-05-27, `Agency_OS-q6ed` P0). When an atom is superseded or deleted, recall must immediately exclude it. Prior state: delete was partial and supersede coverage uneven across the three stores. This PR adds the wiring that fires invalidate-by-memory-id across Hindsight + Weaviate + Valkey on supersede/delete events.

**PR 2 of 4** in the Wave 1+2 Hindsight engineering chain:
1. `Agency_OS-ijf0` — synthesize + trace + complete delete primitives ([#1228](https://github.com/Keiracom/Agency_OS/pull/1228))
2. **this PR** — real-time invalidation hooks
3. `Agency_OS-3rpe` — recency decay function
4. `Agency_OS-3g9t` — atom granularity spec + CI gate

## Changes

`src/keiracom_system/memory/invalidation.py` — three Protocol-typed adapters + coordinator:

- **`HindsightInvalidator`** — calls `DELETE /v1/default/banks/{id}/memories/{id}` on delete-trigger. On supersede-trigger Hindsight is intentionally a no-op (engine preserves history natively; recall filters by the `supersedes` edge written by `AntiPatternWrapper`).
- **`ValkeyInvalidator`** — back-pointer registry. `register_memory_ref(cache_key, memory_ids)` writes a Redis set keyed on memory_id; on invalidate, every cache_key in the set is deleted, then the set itself. Refuses cross-tenant invalidate at the boundary (mirrors `ValkeyClient._enforce_tenant_prefix`).
- **`WeaviateInvalidator`** — delete-by-metadata-filter across configured collections (`Decisions` / `Discoveries` / `Sessions` / `Keis` / `Codebase` by default). Partial-failure case (one collection errors) flags `outcome.ok=False` with `deleted_count` reflecting the surviving collections.
- **`InvalidationCoordinator`** — wires the three adapters and exposes `on_delete(tenant_id, memory_id)` + `on_supersede(tenant_id, memory_id, superseded_by)` hooks. Any adapter may be `None` — early-deploy environments can wire only what's available.

## Fail-test stale-recall pattern

[...]

PR #1228, opened 2026-05-28
[SCOUT] feat(memory): Hindsight synthesize + trace + complete delete primitives (Agency_OS-ijf0)

Keiracom:
## Summary

Closes the **YELLOW-3** gap from Aiden's pre-cutover audit (2026-05-27): 3 of 6 MAL V1 primitives shipped via the 4 domain wrappers (`ingest` / `recall` / `supersede`); the remaining three (`synthesize` / `trace` / `complete_delete`) are cross-cutting query primitives and live in `src/keiracom_system/memory/wrappers/primitives.py` alongside `trace_composition.py`.

This is **PR 1 of 4** in the Wave 1 + 2 Hindsight engineering chain (Dave-ratified 2026-05-27 all-tiers cutover gating, dual-deliberator concur Aiden + Viktor). Sequence:
1. **this PR** — synthesize + trace + complete delete primitives
2. `Agency_OS-q6ed` — real-time invalidation hooks
3. `Agency_OS-3rpe` — recency decay function
4. `Agency_OS-3g9t` — atom granularity spec + CI gate

## Changes

- **`synthesize`** — LLM-grounded reflect with mandatory source-atom pointers. `SynthesisResult.source_atom_ids` is non-empty by construction; zero citations raise `ValueError` ("synthesis drift guard" per Aiden Phase-2 mitigation against ungrounded answers). Handles both `citations` and legacy `cited_memory_ids` field shapes + both `["id"]` and `[{id,content}]` citation shapes.
- **`trace`** — walks supersedes / superseded_by edges around a `memory_id` and returns `ProvenanceChain` with timestamp-ordered `ProvenanceEvent`s. Distinct from `compose_audit_record` (single-reflect `AuditRecord` for one-shot HIPAA/legal/accounting audits).
- **`complete_delete`** — hard-delete via the engine's DELETE endpoint + structured `DeleteRecord` audit-log entry. Closes the partial gap where `delete` was a valid operation in `AuditRecord` but had no method to execute it. Audit invariants (actor + reason mandatory) fail loud.

**Protocol surface change:** `_base.HindsightClient` Protocol extends with `delete(bank_id, memory_id) -> dict`. `FleetHindsightClient` HTTP shim in `keiracom_system/fleet/hindsight/smoke_wrappers.py` implements `DELETE /v1/default/banks/{id}/memories/{memory_id}`.

## Notes — canonical key values (audit-dispatch checklist)

[...]

PR #1234, opened 2026-05-28
[SCOUT] feat(memory): recency decay on retrieval scores — tunable half-life per topology + canonical bypass (Agency_OS-3rpe)

Keiracom:
## Summary

**Wave 2 CUTOVER GATE** (Dave ratify 2026-05-27, `Agency_OS-3rpe` P0). Exponential decay on similarity score by atom age, with half-life keyed off topology and ceo-ratified / canonical-tagged atoms bypassing decay entirely. Solves the named failure mode: recency must not bury a 4-year-old ratified architecture fact under last week's tool-call chatter, but at equal-ish similarity, newer must win.

**PR 3 of 4** in the Wave 1+2 Hindsight engineering chain:
1. `Agency_OS-ijf0` — synthesize + trace + complete delete primitives ([#1228](https://github.com/Keiracom/Agency_OS/pull/1228))
2. `Agency_OS-q6ed` — real-time invalidation hooks ([#1230](https://github.com/Keiracom/Agency_OS/pull/1230))
3. **this PR** — recency decay function
4. `Agency_OS-3g9t` — atom granularity spec + CI gate

## Changes

`src/keiracom_system/memory/recency_decay.py` — pure math + small policy layer; no external clients:

- **`apply_recency_decay(score, age_seconds, half_life_seconds)`** — pure scalar `score * 2^(-age / half_life)`. Treats negative age as zero (clock-skew defensive). Rejects non-positive half-life.
- **`RecencyDecayConfig`** — `half_lives` dict (topology -> days), `default_half_life_days` fallback for unknown topologies, `exempt_tags` set for canonical bypass, `score_floor` clamp so high-similarity old memories don't collapse to zero.
- **`DEFAULT_HALF_LIVES`** — V1 baseline (Dave-ratified starting points; tuning is Phase 2 empirical work):
  - `fleet_keis` 7d, `fleet_agent_memories` 7d, `fleet_tool_calls` 3d, `fleet_session_transcripts` 3d
  - `fleet_discoveries` 14d, `fleet_sessions` 14d
  - `fleet_decisions` 30d
  - `fleet_global_governance_patterns` 60d
  - `fleet_strategic_documents` 90d, `fleet_codebase` 90d
- **`DEFAULT_EXEMPT_TAGS`** = `{canonical, ceo_ratified, ratified, governance_locked}` — atoms carrying any of these tags pass through with score unchanged.
- **`decay_scored_memories(memories, config, now=...)`** — applies policy. Fails loud on missing `id` / `score` / `topology` / wrong tags type — silent fallback would mask wiring bugs at the call site.
- **`ScoredMemory`** dataclass preserves `original_score` so observability + explain-this-rank UX can show the decay delta.
- **`rerank_by_decayed_score`** — stable sort descending.

## Acceptance tests demonstrate the named cutover-gate behaviour

- **Recency flip** (`test_rerank_demonstrates_recency_flip`): old high-similarity (0.95 score, age = 3 half-lives -> 0.125x = 0.119) loses to fresh lower-similarity (0.60 score, age 0). Fresh wins despite being originally less similar.
- **Canonical bypass at high age** (`test_rerank_preserves_canonical_bypass_at_high_age`): 4-year-old `canonical`-tagged atom with score 0.95 still beats fresh chatter at score 0.30.

## Notes — canonical key values (audit-dispatch checklist)

Source: `public.ceo_memory.ceo:cutover_plan_v1` + `public.ceo_memory.ceo:memory_abstraction_layer_v1`.

**`ceo:cutover_plan_v1` — full_retrieval_tier_ratify_2026_05_27_22Z.waves.wave_2_retrieval_core:**
> "hybrid search + cross-encoder reranker + recency decay"

**`ceo:memory_abstraction_layer_v1` — substantive_lock:**
> "mem.topology_tier_keyed" (interpreted here as per-topology half-life tuning)

## Test plan

- [x] 28 tests pass: 6 pure math / 7 decay positive / 5 decay negative / 4 canonical bypass / 3 rerank ordering / 2 constants-table sanity / 1 score-floor
- [x] Pure-math correctness: score halves at exactly age = half_life (`test_apply_recency_decay_halves_score_at_half_life`)
- [x] Per-topology half-life difference verified (`test_decay_respects_per_topology_half_life_difference`)
- [x] All exempt tags tested (`canonical`, `ceo_ratified`, `governance_locked`, custom-replacement)
- [x] Clock-skew defensive case (negative age -> treated as zero)
- [x] `ruff check` clean; `ruff format --check` clean

```
28 passed in 0.13s
```

## Out of scope (intentional)

- **Wiring into `src/retrieval/orchestrator.py` `retrieve_with_outcome`** — the call-site change that applies decay after FlashRank reranking is its own PR (orthogonal scope per `feedback_split_orthogonal_scope`). The signal-attached `ScoredMemory` shape needs a parallel `RetrievedNode` upgrade that touches the agent_query layer — better split.
- **Per-tenant override of `DEFAULT_HALF_LIVES`** — V1 deploys use the module default. Per-tenant override hooks land when tenant-config plumbing is ready.
- **Learned-decay / cohort tuning** — Phase 2 empirical work. Half-life table is config-injected so this lands cleanly later.

### DIFF
--- a/src/keiracom_system/memory/recency_decay.py
+++ b/src/keiracom_system/memory/recency_decay.py
@@ -1,0 +1,268 @@
+"""recency_decay.py — exponential recency decay on retrieval scores.
+
+Wave 2 CUTOVER GATE (Dave ratify 2026-05-27, Agency_OS-3rpe): older memories
+must lose to newer at equal similarity unless they carry a canonical tag
+(ratified ceo_memory keys, governance-locked architecture facts, etc.).
+Pure math + small policy layer; no external clients.
+
+Canonical key citations (per audit-dispatch checklist `_orchestrator.md`):
+
+ceo:cutover_plan_v1 — full_retrieval_tier_ratify_2026_05_27_22Z.waves.wave_2_retrieval_core:
+    "hybrid search + cross-encoder reranker + recency decay"
+
+ceo:memory_abstraction_layer_v1 — substantive_lock:
+    "mem.topology_tier_keyed" (per-topology half-life)
+
+DESIGN — `apply_recency_decay(score, age_seconds, half_life_seconds)` is the
+pure scalar. `decay_scored_memories(memories, *, config, now=...)` applies
+the policy: per-topology half-life lookup + canonical-tag bypass + optional
+floor so decay doesn't collapse a high-similarity old memory below the
+noise threshold.
+
+Half-life table is config-injected, not hardcoded here — different deploys
+tune differently. The DEFAULT_HALF_LIVES dict below is the V1 baseline
+Dave-ratified topology->half-life mapping; production wiring may override
+per-tenant.
+"""
+
+from __future__ import annotations
+
+import logging
+import math
+import time
+from dataclasses import dataclass, field
+from datetime import UTC, datetime
+from typing import Any
+
+log = logging.getLogger(__name__)
+
+
+SECONDS_PER_DAY = 86400
+LN_2 = math.log(2.0)
+
+
+# V1 baseline half-lives per topology (in days). Source: Dave ratify
+# 2026-05-27 (ceo:cutover_plan_v1 wave_2_retrieval_core) — values chosen so
+# fast-moving banks (KEI tracker, agent memories) decay quickly and
+# slow-moving banks (codebase, decisions) decay slowly. Tuning is empirical
+# Phase 2 work; these are starting points, not final.
+DEFAULT_HALF_LIVES: dict[str, float] = {
+    "fleet_keis": 7.0,
+    "fleet_agent_memories": 7.0,
+    "fleet_tool_calls": 3.0,
+    "fleet_session_transcripts": 3.0,
+    "fleet_discoveries": 14.0,
+    "fleet_decisions": 30.0,
+    "fleet_global_governance_patterns": 60.0,
+    "fleet_strategic_documents": 90.0,
+    "fleet_codebase": 90.0,
+    "fleet_sessions": 14.0,
+}
+
+# Tags that EXEMPT a memory from decay. Ratified ceo_memory keys + canonical
+# architecture facts must always surface — recency must not bury "Hindsight
+# was chosen as MAL engine 2026-05-24" under last week's tool-call chatter.
+DEFAULT_EXEMPT_TAGS: frozenset[str] = frozenset(
+    {"canonical", "ceo_ratified", "ratified", "governance_locked"}
+)
+
+
+@dataclass(frozen=True)
+class RecencyDecayConfig:
+    """Per-deploy recency-decay policy.
+
+    `half_lives` maps topology key -> half-life in days. Unknown topology
+    falls back to `default_half_life_days` (None = no decay applied).
+    `exempt_tags` short-circuits decay when any of these tags are present
+    on the memory.
+    `score_floor` clamps the decayed score from below so high-similarity
+    old memories don't collapse to zero (e.g. a 4-year-old decision with
+    similarity 0.95 still beats a 1-hour-old chatter with 0.10).
+    """
+
+    half_lives: dict[str, float] = field(default_factory=lambda: dict(DEFAULT_HALF_LIVES))
+    default_half_life_days: float | None = 30.0
+    exempt_tags: frozenset[str] = DEFAULT_EXEMPT_TAGS
+    score_floor: float = 0.0
+
+
+@dataclass(frozen=True)
+class ScoredMemory:
+    """Scored retrieval candidate.
+
+    `original_score` preserves the pre-decay similarity so callers can
+    inspect the delta (e.g. observability + explain-this-rank UX).
+    """
+
+    memory_id: str
+    score: float
+    original_score: float
+    age_seconds: float
+    topology: str
+    decay_applied: bool
+    exempt_reason: str = ""
+    metadata: dict[str, Any] = field(default_factory=dict)
+
+
+def apply_recency_decay(score: float, age_seconds: float, half_life_seconds: float) -> float:
+    """Pure exponential decay: score * 2^(-age / half_life).
+
+    Equivalent to `score * exp(-ln(2) * age / half_life)`. Half-life is the
+    age at which a score halves; the curve is monotonically decreasing in age.
+    """
+    if half_life_seconds <= 0:
+        raise ValueError(f"half_life_seconds must be positive, got {half_life_seconds!r}")
+    if age_seconds < 0:
+        # A negative age (future-dated atom) is a clock-skew signal — log
+        # but do not penalise; treat as age=0.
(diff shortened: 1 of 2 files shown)


Keiracom:
[MAX] REVIEW — PR #1234 feat(memory): recency decay on retrieval scores (Agency_OS-3rpe)
[REVIEW:approve:max]

**Lens:** code quality + test coverage (not implementation feasibility).

---

**Summary**

Solid pure-math + policy layer. Good separation: `apply_recency_decay` is a side-effect-free scalar, `decay_scored_memories` is the policy loop, `rerank_by_decayed_score` is a one-liner sort. Test suite (394 lines) states its coverage matrix explicitly and delivers on it. No blocking issues.

---

**Findings**

| # | Severity | Location | Finding |
|---|----------|----------|---------|
| 1 | medium | `recency_decay.py` L~194 / `test_recency_decay.py` | Missing negative test: `decay_scored_memories` when `score` is non-numeric (e.g. `"high"`). Code does `original_score = float(score)` which raises bare `ValueError: could not convert string to float` — no guard, no `RerankClientError`-equivalent wrapper. Caller gets an undocumented exception type. Add a type-check + descriptive error matching the other validation blocks, and a test. |
| 2 | low | `recency_decay.py` L~130 | `apply_recency_decay` has no guard for `age_seconds = float("nan")` or `float("inf")`. Both propagate silently into `ScoredMemory.score` (`nan * factor = nan`; `inf * factor = inf`). Upstream shouldn't produce these but a defensive `math.isfinite` check + test would prevent score corruption if clock-skew or a malformed payload slips through. |
| 3 | low | `recency_decay.py` L~56 | `DEFAULT_HALF_LIVES` is a plain `dict` at module level. `RecencyDecayConfig` correctly copies it via `default_factory`, but a module-level mutation (`DEFAULT_HALF_LIVES["fleet_keis"] = 999`) would affect all subsequent `RecencyDecayConfig()` calls. `MappingProxyType` or `frozenset`-equivalent immutability (already used for `DEFAULT_EXEMPT_TAGS`) would close this. Nit. |
| 4 | low | `recency_decay.py` L~150 | `rstrip("Z")` strips the trailing character set `{"Z"}` not a literal suffix. For the well-formed ISO-8601 strings TEI produces this is fine, but a malformed string like `"2026-01-01T00:00:00ZZZ"` silently becomes `"2026-01-01T00:00:00"` and parses as midnight 2026-01-01 instead of returning `None`. Consider `removesuffix("Z")` (Python 3.9+) which only removes if the full suffix matches. |
| 5 | nit | `test_recency_decay.py` | No explicit test for the default `score_floor=0.0` case (that floor is not applied when default config is used and decay produces a value above zero). Covered implicitly by `test_decay_decays_score_per_topology_half_life` but a dedicated assertion would document intent. |

---

**Verdict**

No blocker. Item 1 (medium) is the only item worth a follow-up: the undocumented `ValueError` on non-numeric score is a correctness gap that would produce a cryptic traceback in production. Items 2–5 are low/nit. The core decay math tests are deterministic (fixed `_NOW` epoch, no I/O, no sleeps), the canonical-bypass tests cover all four exempt tags, and the recency-flip acceptance test is a strong regression guard for the cutover-gate requirement.

Recommend: fix item 1 in-PR (GOV-10 resolve-now-not-later), acknowledge items 2–5 as follow-up if not addressed. Merge-eligible after item 1 resolved or author explicitly acknowledges the exception-type gap.

[MAX]

Keiracom:
[ELLIOT] Implementation-feasibility lens review — PR #1234: [SCOUT] feat(memory): recency decay on retrieval scores — tunable half-life per topology + canonical bypass (Agency_OS-3rpe)

[REVIEW:approve:elliot]

## Implementation assessment

- **Module is correctly isolated — pure math, no external clients.** `recency_decay.py` has zero network calls, no Supabase imports, no Hindsight HTTP. The decay function (`score * 2^(-age/half_life)`) is a closed-form scalar. The deliberate non-wiring into `retrieve_with_outcome` is correct: the PR description flags this as intentional scope split per `feedback_split_orthogonal_scope`. Integrating here would have forced a `RetrievedNode` shape change touching the agent_query layer — that is correctly deferred.

- **Canonical-tag bypass prevents the named failure mode at runtime.** A 4-year-old `canonical`-tagged atom does not decay. The tag check is a simple frozenset intersection so it adds negligible latency per memory. The config is per-deploy injectable meaning per-tenant override of exempt tags is already structurally supported when that plumbing lands.

- **Clock-skew handling is defensive but not silent.** Negative `age_seconds` logs a warning and treats as age=0 (no penalty). Missing `created_at` / `timestamp` falls through to no decay with `exempt_reason="no_timestamp"` — this is observable and won't silently suppress decay for a whole topology if the upstream payload is misshapen. The fail-loud approach on missing `id`/`score`/`topology` is the right call here.

- **One wiring gap to flag (non-blocking for this PR, per stated scope).** Until the follow-on PR wires `decay_scored_memories` into `retrieve_with_outcome`, the decay logic is deployed but inert in production. That is the intended state for this PR — the cutover-gate behaviour exists in tests but not in the live retrieval path. The follow-on PR that does the wiring will be the real feasibility gate for the recency-flip guarantee. Flag for Aiden's architecture review: ensure the wiring PR handles the `RetrievedNode → ScoredMemory` shape conversion without a silent no-op if `created_at` is absent in the existing node payload.

- **28 tests pass in 0.13s; pure-unit, no I/O.** Coverage includes the two acceptance criteria (recency flip and canonical preservation), all exempt tags, clock-skew case, score-floor clamp, and the constants regression guard. The topology coverage test (`test_default_half_lives_covers_known_topologies`) is a good catch — will fail loudly if a new bank topology is added without a corresponding half-life entry.

Keiracom:

The canonical bypass is the key governance guarantee I was checking: atoms tagged {canonical, ceo_ratified, ratified, governance_locked} pass through with score unchanged. A 4-year-old ratified architecture decision will never be buried under last week's session chatter regardless of half-life. This directly upholds the MAL V1 principle that ratified state must remain surfaceable.

Three other things I verified:

1. **Timestamp timezone handling** — _parse_atom_timestamp() strips 'Z', assumes UTC for naive datetimes, handles epoch int/float. On parse failure, returns None → age_seconds = 0.0 (treated as fresh, no decay). This is the correct fail-open: a timestamp parse bug should not corrupt retrieval scores.

2. **Half-life table is config-injected, not hardcoded** — per-tenant override is possible. The DEFAULT_HALF_LIVES table is Dave-ratified V1 baselines (noted as Phase 2 empirical tuning starting points). The decay applies to the topology (bank), not the memory type — this is the right granularity for the current Weaviate collection structure.

3. **score_floor** — prevents high-similarity old memories from collapsing to zero. Default 0.0 means no floor unless configured. Operators can tune this per-deploy without code change.

Pure math + policy layer with no external clients is the correct scope for this primitive — testable in isolation without service dependencies.

[REVIEW:approve:aiden]

PR #1225, opened 2026-05-27
[ORION] feat(dispatcher-wiring-KEI213): spawn attribution + per-task-type wired into /dispatcher/spawn (Agency_OS-r9p3 PR D)

Keiracom:
## Summary

Cutover step 4.5 dispatcher-wiring PR **D** for **KEI-213** — wires PR #1207 spawn-attribution + PR #1209 per-task-type telemetry into `/dispatcher/spawn` per Dave + Aiden + Viktor dual-concur 2026-05-27.

**Sibling KEI-213 PRs (parallel pipeline):**
- PR #1222 (A) — idempotency gate
- PR #1223 (B) — budget ceiling gate
- PR #1224 (C) — context-window gate (interceptor_proxy)
- This PR (D) — spawn attribution + per-task-type
- (Coming) E — wiring inventory + CI guard update

## Changes

| File | Change | LoC |
|---|---|---|
| `src/dispatcher/main.py` | MOD — imports attribution module; `attribution_enabled` toggle (env); helper fns; `_emit_attribution` call after `_register_session` | +70 |
| `tests/dispatcher/test_main_spawn_attribution_wiring.py` | NEW — 27 unit tests | +207 |

## Wiring point

```
/dispatcher/spawn (HTTP POST)
  ↓ idempotency gate (PR A) — if wired
  ↓ budget ceiling gate (PR B) — if wired
  ↓ backend validation
  ↓ manager.spawn(**spawn_kwargs)
  ↓ _register_session(...)
  ↓ [NEW] _emit_attribution(...)        ← SpawnAttributionEntry JSONL written here
  ↓ HTTP 200 response
```

Emit fires AFTER successful register so only sessions that actually entered supervision get telemetry rows.

## Source_type derivation (PR #1207 SOURCE_TYPES)

| Trigger | source_type |
|---|---|
| `spawn_kwargs.source_type` validated against frozenset | (as given) |
| `spawn_kwargs.from == "dave"` | `slack` |
| `spawn_kwargs.from in {"cron", "scheduler"}` | `cron` |
| Default | `inbox` (KEI-213 dispatcher's primary dispatch path) |

`unknown` reserved for explicit envelope ambiguity per PR #1207's "silent fallback is a bug" contract.

## Task_type derivation (PR #1207/#1209 TASK_TYPES)

| Trigger | task_type |
|---|---|
| `spawn_kwargs.task_type` validated | (as given) |
[...]

PR #1223, opened 2026-05-27
[ORION] feat(dispatcher-wiring-KEI213): budget ceiling gate wired into /dispatcher/spawn (Agency_OS-r9p3 PR B)

Keiracom:
## Summary

Cutover step 4.5 dispatcher-wiring PR **B** for **KEI-213** — wires `BudgetCeilingGate` (PR #1203 / Cat 21 lever 28 / cutover-blocker 2) into the canonical `/dispatcher/spawn` endpoint.

**Sibling KEI-213 PRs (parallel pipeline):**
- PR #1222 (A) — idempotency gate
- This PR (B) — budget ceiling gate
- (Coming) C — context-window gate (into interceptor_proxy)
- (Coming) D — spawn attribution + per-task-type
- (Coming) E — wiring inventory + CI guard update

## Changes

| File | Change | LoC |
|---|---|---|
| `src/dispatcher/main.py` | MOD — imports BudgetCeilingGate; module-level `_budget_gate` + `_set_budget_gate()` DI; `_BUDGET_PROCEED` frozenset; gate call in `/dispatcher/spawn` | +44 |
| `tests/dispatcher/test_main_budget_wiring.py` | NEW — 5 unit tests | +206 |

## Decision → action mapping

| BudgetDecision | Action |
|---|---|
| `SPAWN_OK` | proceed to manager.spawn() |
| `OVERAGE_LOG_AND_SPAWN` | proceed (priority over budget) |
| `DAVE_BYPASS` | proceed (CEO never blocked) |
| `FORCE_OVERRIDE` | proceed (--force-spawn) |
| `QUEUE_NEXT_DAY` | HTTP 200, no spawn, decision in body |
| `DROP_WITH_LOG` | HTTP 200, no spawn, decision in body |

## Priority + source derivation

```python
sk = req.spawn_kwargs or {}
priority = sk.get("priority")  # high/normal/low (validated; fallback NORMAL)
source = sk.get("source")  # validated against SOURCE_DAVE_DM/FLEET
# OR derived from sk.get("from") — "dave" → SOURCE_DAVE_DM (Haiku bypass)
```

## Response shape on skip-spawn

HTTP 200 (decision is **expected behaviour** — budget gate fired, not an error):

```json
{
  "spawned": false,
  "key": "test-key-1",
  "backend": "tmux",
  "decision": "queue_next_day",
  "current_day_spend_aud": 79.0,
  "daily_budget_aud": 25.0,
  "reason": "normal-priority over budget — defer to next day"
}
[...]

PR #1237, opened 2026-05-28
[ORION] fix(ci): update hook-assertion tests after Dave hook-kill directive 2026-05-27

Keiracom:
## Summary

Three settings.json wiring smoke tests asserted the presence of hook entries in `.claude/settings.json`. Dave directive 2026-05-27 removed all Claude Code hooks; the file no longer carries a `hooks` key, so these assertions cannot pass and break Backend Tests on every PR.

Marked each with `@pytest.mark.xfail` referencing the directive:

- `tests/scripts/hooks/test_env_schema_validate.py::test_hook_registered_first_in_session_start_chain`
- `tests/scripts/test_governance_hooks.py::test_settings_json_contains_pretooluse_hook`
- `tests/scripts/test_session_end_hook.py::test_settings_json_contains_session_end_hook`

The underlying scripts (`env_schema_validate.sh`, `governance_hooks.py`, `session_end_hook.py`) still exist and are exercised by the other tests in each file — only the "is this wired into settings.json" smoke is xfailed.

## Test plan

- [x] Run the 3 affected tests in isolation: 3 xfailed ✓
- [x] Run all 3 affected files end-to-end: 64 passed, 3 xfailed in 0.46s ✓
- [ ] CI Backend Tests passes on this PR

## Out of scope

- Production code (no changes)
- `.claude/settings.json` (no changes — Dave's hook-kill directive owns that file)

### DIFF
--- a/tests/scripts/hooks/test_env_schema_validate.py
+++ b/tests/scripts/hooks/test_env_schema_validate.py
@@ -176,6 +176,11 @@ def test_script_syntax_valid() -> None:
 # ---------------------------------------------------------------------------
 
 
+@pytest.mark.xfail(
+    reason="Claude Code hooks removed per Dave directive 2026-05-27 (hook-kill); "
+    ".claude/settings.json no longer carries a 'hooks' key. The env_schema_validate.sh "
+    "script still exists and its behaviour is exercised by the other tests in this file."
+)
 def test_hook_registered_first_in_session_start_chain() -> None:
     settings = json.loads((REPO_ROOT / ".claude" / "settings.json").read_text())
     session_start = settings["hooks"]["SessionStart"]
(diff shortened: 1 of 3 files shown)


Keiracom:

xfail (not delete) is the correct treatment here. The underlying scripts still exist and are tested by the other cases in each file — only the settings.json wiring smoke is suspended. If hooks are reinstated, xfail becomes xpass, which surfaces prominently in test output and prompts removal of the markers. That feedback loop is correct.

The xfail reason strings cite the directive explicitly ('Dave directive 2026-05-27 hook-kill') so future readers know this is not technical debt to clean up but a deliberate governance change. Governance changes need a traceable reason in the test, not a bare skip.

strict=False (default) is the right xfail mode here — strict=True would turn a future xpass into a failure, which is too aggressive for a configuration-driven removal that could be reversed.

[REVIEW:approve:aiden]

Keiracom:
[ELLIOT] Closing duplicate — PR #1238 opened by Scout covers the same fix with the preferred Option A (inverted compliance gates). Orion please focus on rebasing #1223 + #1225.

PR #1239, opened 2026-05-28
[NOVA] fix(wave12-gate): 4 blocking gaps from Max re-audit

Keiracom:
## Summary
Fixes all 4 GATE-HOLD blocking gaps from Max's Wave 1+2 re-audit, in a single PR. Wave 3 is blocked until these clear CI. No logic changes — two Go tests, one systemd sandbox path, one default-URL constant + its test.

| # | Gap | Type | Fix |
|---|-----|------|-----|
| 1 | go_sidecar `/proxy` malformed-JSON → 400 | missing test | `TestProxy_MalformedJSON400` sends invalid JSON, asserts 400 |
| 2 | go_sidecar `/proxy` rate-limit → 429 | missing test | `TestProxy_RateLimit429` empties the per-tenant token bucket on the proxy hot path (not just `/validate`), asserts 429 |
| 3 | go_sidecar systemd `ExecStartPre` EROFS | runtime bug | `ProtectHome=read-only` blocked `build.sh` writing `bin/sidecar` on fresh deploy → add `ReadWritePaths=%h/clawd/Agency_OS/infra/keiracom_system/go_sidecar/bin` |
| 4 | reranker `DEFAULT_BASE_URL` DNS-fails for host callers | runtime bug | `http://reranker:80` only resolves inside Docker; host-process callers (Hindsight) fail silently → switch to `http://localhost:8090` (TEI host port) + test that `healthy()` logs a WARNING (not silent) when unreachable |

## Verification (raw)
```
go test ./...                          → ok (9 proxy/validate tests, incl. 2 new)
pytest tests/keiracom_system/reranker  → 21 passed, 1 skipped
ruff format --check                    → 2 files already formatted
ruff check                             → All checks passed!
```

Note: gap-4 also updates the existing `test_default_base_url_constant` assertion to match the new constant (would otherwise fail).

## Reviewer ask
Max to re-review and lift GATE-HOLD on Wave 3.

### DIFF
--- a/infra/keiracom_system/go_sidecar/cmd/sidecar/proxy_test.go
+++ b/infra/keiracom_system/go_sidecar/cmd/sidecar/proxy_test.go
@@ -176,4 +176,45 @@ func TestProxy_UnknownTenant(t *testing.T) {
 	}
 }
 
+// 8. /proxy: malformed JSON body → 400.
+func TestProxy_MalformedJSON400(t *testing.T) {
+	cfg := fixtureCfg("http://unused")
+	v, rl, bm, hc := setupSidecar(cfg)
+	h := proxyHandler(cfg, v, rl, bm, hc)
+	rr := httptest.NewRecorder()
+	h.ServeHTTP(rr, httptest.NewRequest("POST", "/proxy", strings.NewReader("{not valid json")))
+	if rr.Code != http.StatusBadRequest {
+		t.Fatalf("expected 400 on malformed JSON; got %d body=%s", rr.Code, rr.Body.String())
+	}
+}
+
+// 9. /proxy: per-tenant rate limiter returns 429 once bucket empty (mirrors the
+// /validate rate-limit test but exercises the limiter on the proxy hot path).
+func TestProxy_RateLimit429(t *testing.T) {
+	upstream := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		w.WriteHeader(200)
+		_, _ = w.Write([]byte(`{"result":"ok"}`))
+	}))
+	defer upstream.Close()
+	cfg := fixtureCfg(upstream.URL)
+	tA := cfg.Tenants["tenant_a"]
+	tA.RateLimit = ratelimit.Spec{RPS: 0.0001, Burst: 1}
+	cfg.Tenants["tenant_a"] = tA
+	v, rl, bm, hc := setupSidecar(cfg)
+	h := proxyHandler(cfg, v, rl, bm, hc)
+	body, _ := json.Marshal(proxyRequest{TenantID: "tenant_a", Tool: "read_file", Server: "search", Body: json.RawMessage(`{"q":"hi"}`)})
+	// first call consumes the single token and forwards to upstream
+	rr := httptest.NewRecorder()
+	h.ServeHTTP(rr, httptest.NewRequest("POST", "/proxy", bytes.NewReader(body)))
+	if rr.Code != 200 {
+		t.Fatalf("first call expected 200; got %d body=%s", rr.Code, rr.Body.String())
+	}
+	// second call — bucket empty
+	rr = httptest.NewRecorder()
+	h.ServeHTTP(rr, httptest.NewRequest("POST", "/proxy", bytes.NewReader(body)))
+	if rr.Code != http.StatusTooManyRequests {
+		t.Fatalf("second call expected 429; got %d body=%s", rr.Code, rr.Body.String())
+	}
+}
+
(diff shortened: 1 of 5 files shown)


Keiracom:
[MAX] Code quality + test coverage lens review — PR #1239: Wave 1+2 gate-hold fixes.

[REVIEW:approve:max]

Verified all 4 blocking gaps from my re-audit are correctly addressed.

**Gap 1 — /proxy malformed-JSON → 400 (CLOSED)**
`TestProxy_MalformedJSON400` sends `"{not valid json"` through `proxyHandler` and asserts `http.StatusBadRequest`. Correct path, correct assertion. The security enforcement boundary's error-path behaviour is now locked.

**Gap 2 — /proxy rate-limit → 429 end-to-end (CLOSED)**
`TestProxy_RateLimit429` wires a real upstream stub, sets `Burst:1 / RPS:0.0001` (no meaningful refill window), fires two requests through `proxyHandler` — first returns 200 (consumes the token), second returns 429 (bucket empty). Genuine end-to-end through the proxy handler, not just the /validate path. Correct.

**Gap 3 — systemd EROFS on fresh deploy (CLOSED)**
Two-part fix both present: (a) `ReadWritePaths=%h/clawd/.../go_sidecar/bin` added to the unit file — `%h` is the correct systemd specifier for $HOME in user units; (b) `install-systemd.sh` runs `mkdir -p "${ROOT}/bin"` before `daemon-reload` — systemd refuses to start a unit whose `ReadWritePaths` target is absent, and `bin/` is gitignored so it wouldn't exist on a fresh checkout without this. Both halves required, both present.

**Gap 4 — DEFAULT_BASE_URL docker-DNS silent failure (CLOSED)**
Changed to `"http://localhost:8090"` with an inline comment explaining the docker-network vs host-process distinction. Constant test updated. Bonus: `test_healthy_unreachable_is_not_silent` added — verifies that `healthy()=False` on a connection error also emits a WARNING log, giving operators a signal rather than silent degradation. Exceeds the original ask.

**Remaining nit (non-blocking, carry-forward):** `var _ = io.EOF` at the bottom of proxy_test.go is still an unused import keeper — not introduced by this PR, not worth blocking on.

GATE-HOLD lifted. Wave 3 is clear from Max lens.

Keiracom:
[ELLIOT] Implementation-feasibility lens review — PR #1239: [NOVA] fix(wave12-gate): 4 blocking gaps from Max re-audit

[REVIEW:approve:elliot]

## Implementation assessment

- **Gap 1 — malformed-JSON → 400 (test only):** Production path confirmed at `main.go:102-104` — `json.NewDecoder(r.Body).Decode` already returns 400 on decode error. The test sends `{not valid json`, calls `proxyHandler` directly via `httptest`, and asserts `rr.Code == 400`. Test exercises the real code path. No dead stub risk.

- **Gap 2 — rate-limit → 429 (test only):** `TestProxy_RateLimit429` sets `RPS: 0.0001, Burst: 1` on tenant_a, fires two sequential requests through the same handler instance, asserts first → 200, second → 429. Production path at `main.go:106-108` calls `rl.Allow(req.TenantID)` before validation — the test exercises that exact branch on the proxy hot path (distinct from the validate-endpoint test). Mechanically sound; deterministic because no goroutine concurrency and bucket state is not shared across test cases.

- **Gap 3 — systemd EROFS (runtime bug):** `ProtectHome=read-only` + `ReadWritePaths=%h/clawd/Agency_OS/infra/keiracom_system/go_sidecar/bin` is the correct fix — `%h` expands to the service user's home at load time. The companion `mkdir -p ${ROOT}/bin` in `install-systemd.sh` is necessary because systemd refuses to start a unit whose `ReadWritePaths` target is absent; `bin/` is gitignored so it won't exist on a fresh checkout. Both halves of the fix are present and ordered correctly (mkdir in install script runs before daemon-reload).

- **Gap 4 — reranker DEFAULT_BASE_URL (runtime bug):** Switch from `http://reranker:80` (Docker-internal DNS only) to `http://localhost:8090` is correct for host-process callers (Hindsight runs on the host). The `healthy()` method already had a `log.warning` on exception (`reranker_client.py:113`) before this PR; the new `test_healthy_unreachable_is_not_silent` test validates that the warning is emitted with the error message in the log record — closes the silent-failure gap. The existing constant test updated to match.

- **One nit — not a hold:** The rate-limit test creates a live upstream `httptest.Server` but the first call only gets to 200 if `v.Allow` passes for `tenant_a`. The fixture config must have `tenant_a` defined with the `read_file` tool and `search` server whitelisted — confirmed by the existing `fixtureCfg` usage in the adjacent tests. No issue, just noting the dependency is implicit.

PR #1241, opened 2026-05-28
[ORION] docs(retrieval): FlashRank characterisation baseline (pre-Wave-3 Option B)

Keiracom:
## Summary

Pre-Wave-3 baseline for the retrieval reranker swap (FlashRank → Nova cross-encoder sidecar). **Option B** (ground-truth-free characterisation) per Aiden's decision after Orion's directive-scrutiny GAP report ruled out precision@5 as indefensible (no labelled golden set, FlashRank not installed locally, premise inverted).

Adds:
- `docs/retrieval/flashrank_baseline_2026-05-28.md` — the artifact Wave 3 diffs against
- `scripts/retrieval/flashrank_characterization.py` — re-runnable harness (ruff clean)

Measured over 10 governance/decisions queries against live Hindsight: rank churn, score distributions, rerank latency. **Explicitly NOT precision@5** — characterisation of reorder behaviour + cost only, no quality claim.

## Key findings

- **8 of 10 fleet Hindsight banks are empty** (only `global_governance_patterns`=19, `decisions`=2). Any retrieval-quality A/B is currently on a near-empty corpus — flagged as an indexing follow-up.
- **At the production-default 200ms rerank budget, FlashRank bypasses on every query** (measured median 336ms on a 21-doc pool; run forced with 5000ms budget to characterise the rerank path). Currently-shipped default behaviour is effectively raw ANN.
- **FlashRank meaningfully reorders**: mean 2.1/5 displaced, top-1 changed in 4/10.

## Scope

- Read-only against Hindsight; **no production code paths touched**.
- `flashrank==0.2.10` installed into the venv to exercise the reranker (reversible).
- Option A (labelled golden query set + precision@K framework) filed as bd `Agency_OS-mz2s`, assigned to scout — does not block Wave 3.

## Test plan

- [x] Harness runs end-to-end against live Hindsight (10/10 queries reranked at 5000ms budget)
- [x] `ruff format --check` + `ruff check` clean on the harness
- [ ] Deliberator review of methodology + caveats

### DIFF
--- a/docs/retrieval/flashrank_baseline_2026-05-28.md
+++ b/docs/retrieval/flashrank_baseline_2026-05-28.md
@@ -1,0 +1,149 @@
+# FlashRank characterisation baseline — NOT precision@5
+
+**Date:** 2026-05-28
+**Author:** ORION (Aiden's build clone)
+**main SHA at measurement:** `7675c2355b2018fdc25ff3f0798bb8403b549b4a`
+**Dispatch:** Elliot → Aiden → Orion. Option B (ground-truth-free proxy) per Aiden's decision on Orion's GAP report.
+
+> ⚠️ **This is a characterisation of reranker *behaviour and cost*, not a quality metric.**
+> It deliberately does **NOT** report precision@5. There is no labelled golden query
+> set and no ground-truth relevance judgments, so a precision number measured now
+> would be indefensible as a ship signal. The proper labelled eval harness is filed
+> separately as the Option-A bd issue. These numbers say *how much FlashRank reorders
+> the raw ANN candidates and how long it takes* — they say nothing about whether the
+> reordered order is **better**.
+
+## Why Option B (and not precision@5)
+
+Orion's directive-scrutiny pass found three blocking gaps for a precision@5 baseline:
+
+1. **FlashRank was not installed** in this worktree venv — every query would have
+   bypassed to raw ANN (`reason="flashrank_not_available"`, `rerankers.py:56-59`),
+   producing a raw-ANN number mislabelled as FlashRank. *Resolved for this run* by
+   installing `flashrank==0.2.10` into the venv (reversible; no production code touched).
+2. **The premise was inverted** — FlashRank is *already* the live reranker
+   (`orchestrator.py:7-8` → `rerankers.rerank_top_k`). The Nova cross-encoder sidecar
+   Wave 1 (#1229) and Wave 2 (#1232) already merged to main. "Wave 3" *swaps*
+   FlashRank → sidecar; it does not introduce reranking.
+3. **No ground truth** — no canonical eval query set exists; one agent hand-labelling
+   its own queries is subjective and non-reproducible.
+
+Aiden chose Option B: characterise the current FlashRank reorder behaviour as the
+pre-swap artifact Wave 3 diffs against.
+
+## Method
+
+- **Harness:** `scripts/retrieval/flashrank_characterization.py` (committed alongside this
+  report — re-runnable for the Wave-3 after-picture).
+- **Backend:** live Hindsight at `http://localhost:8889`, real `/memories/recall` per
+  collection (read-only; no writes).
+- **Reranker:** in-process FlashRank, model `ms-marco-MiniLM-L-12-v2`, `flashrank==0.2.10`,
+  Python 3.12.3.
+- **Params:** `k_initial=20`, `k_returned=5` (orchestrator defaults).
+- **Latency budget:** raised to `5000ms` for this run via `AGENCY_OS_RERANK_BUDGET_MS`
+  to *force the rerank path* (see Finding 2 — at the production default it would bypass).
+- **Per query, measured:** rank churn (raw-ANN top-5 identities surviving into the
+  FlashRank top-5, plus mean absolute position shift), raw-ANN scores vs FlashRank
+  relevance scores for the top-5, and wall-clock ms around `rerankers.rerank_top_k`.
+- **Node identity:** `metadata.chunk_id` (falls back to `external_id`, then text hash).
+
+### Collections queried
+
+Probed all ten fleet banks first. **Eight of ten are empty** (see Finding 1), so the
+harness queries only the two populated banks:
+
+- `fleet_global_governance_patterns` — 19 documents
+- `fleet_decisions` — 2 documents
+
+Combined pool = 21 candidates per query (recall over a small corpus returns the whole
+corpus, so each query reorders the same 21 docs — the ANN *order* differs per query).
+
+### Query set (10, governance/decisions-themed to hit the populated banks)
+
+```
+1.  agent governance rules and laws
+2.  Step 0 RESTATE requirement before execution
+3.  how are pull requests reviewed and merged
+4.  callsign discipline and identity enforcement
+5.  FlashRank cross-encoder reranker design
+6.  Hindsight memory recall reader cutover
+7.  dispatcher spawn budget ceiling gate
+8.  bounded spawn one task per spawn enforcement
+9.  Linear and Beads issue tracker sync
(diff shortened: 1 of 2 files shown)


Keiracom:

Methodology integrity is sound. Explicit NOT-precision@5 labelling throughout, clear methodology section, re-runnable harness, correct churn calculation (effective_k boundary handling for pool < k_returned). The artifact does what Option B promised.

Three architecture observations for the record:

**1. The two findings are more valuable than the numbers.** Finding 2 (200ms budget = always bypassed in production) is the architecturally significant result here. The characterisation table says FlashRank meaningfully reorders — but that reordering never fires at the production default. Wave 3 is not 'FlashRank vs sidecar' in the current production state; it is 'raw ANN vs sidecar' because the current reranker is effectively disabled. Wave 3 framing and the A/B comparison baseline must reflect this reality.

**2. Wave 3 has two blockers before a meaningful A/B, not one.** (a) Budget: the Nova sidecar needs to fit inside a budget where FlashRank does not — if the sidecar also runs at 300ms+ it will also bypass, making Wave 3 a no-op. Latency gate must be part of the Wave 3 acceptance criteria. (b) Corpus: 21 docs is too thin for a retrieval quality signal. Finding 1 (8/10 banks empty) means any quality A/B is on noise. Indexing pipeline health must be resolved before the A/B comparison is a ship signal.

**3. Minor harness note:**  is a private method (underscore prefix). Fine for a measurement script — noting so nobody assumes it's a stable API surface. If orchestrator.py is refactored, the harness breaks silently.

[REVIEW:approve:aiden]

Keiracom:
[ELLIOT] Implementation-feasibility lens review — PR #1241: [ORION] docs(retrieval): FlashRank characterisation baseline (pre-swap)
[REVIEW:approve:elliot]

## Implementation assessment

- **No production code paths touched.** Both added files are a doc artifact and a re-runnable research script — zero risk of regression to the live retrieval stack. flashrank==0.2.10 installed into venv is noted as reversible.
- **Harness imports cleanly against the existing stack.** `scripts/retrieval/flashrank_characterization.py` imports `src.retrieval.orchestrator` and `src.retrieval.rerankers` — both present on main. Constants reused from their canonical homes (`DEFAULT_K_INITIAL`, `DEFAULT_K_RETURNED`, `FLEET_TENANT_SLUG`), so the harness stays aligned if those defaults change.
- **Bypass-at-production-budget finding is a genuine runtime risk.** Measured median 336ms against a 21-doc pool; default budget is 200ms. This means Wave 3 A/B will be comparing sidecar against raw-ANN (not FlashRank) at production defaults. The doc calls this out clearly as a tuning-knob concern — Wave 3 authors need to revisit `DEFAULT_BUDGET_MS` before treating any A/B as meaningful.
- **Empty-corpus finding limits Wave 3 validity.** 8/10 fleet banks empty at measurement time. The doc flags this as the most important caveat; the follow-up indexing issue is noted. No implementation action required here, but Wave 3 should gate its quality comparisons on corpus fullness, not just reranker swap.
- **ruff clean and methodology is sound.** Option B (ground-truth-free) is the defensible choice given no labelled set — precision@5 on unlabelled data would be non-reproducible. The churn and latency metrics are computable without labels and are exactly what a pre-swap baseline needs.

Keiracom:
[MAX] Code quality + test coverage lens — medium effort review.

[REVIEW:HOLD:max]

**Findings:**

**1. [medium] `_churn()` effective_k uses `max` where it should use `min`**
`effective_k = max(len(raw_top5), len(rerank_top5))` — if pool < k_returned both lists are shorter than 5. Taking `max` makes `displaced_from_topk = max - survivors`, which over-counts displacement against slots that neither list fills. Should be `min(len(raw_top5), len(rerank_top5))` (the actual shared top-k window). This fires on the `fleet_decisions` pool of 2 docs, where both lists have 2 entries, not 5.

**2. [medium] No tests for the re-runnable harness**
`flashrank_characterization.py` is explicitly described as "re-runnable for the Wave-3 after-picture" — making it a long-lived measurement artefact, not a throwaway script. Both `_churn()` and `_node_key()` are pure functions; a test of each (including the edge case of pool < k_returned that triggered finding #1) would be trivially written and would catch the effective_k bug. Route to test-4 before merge, or bundle a `tests/retrieval/test_flashrank_characterization.py` with fixes.

**3. [low] Private API coupling in `_flashrank_scores()`**
Calls `rerankers._get_reranker()` and `rerankers._to_passage_dicts()` directly (underscore-prefixed private helpers). Aiden already flagged `orchestrator._gather_ann_pool` in the same script. For a measurement-only harness this is tolerable, but if orchestrator.py is refactored the harness breaks silently with no test to catch it. A comment at the top of the function noting these are internal hooks would at minimum surface the coupling explicitly.

**4. [low] Deferred import inside `_flashrank_scores()`**
`from flashrank import RerankRequest` is nested inside the function body. The established pattern in `rerankers.py` is a module-level guarded import. Minor consistency nit.

**5. [nit] `COLLECTIONS` names diverge from doc**
Script uses `("Global_governance_patterns", "Decisions")`; the baseline doc names them `fleet_global_governance_patterns` and `fleet_decisions`. If these resolve identically via Hindsight aliasing, add a comment; if not, one is wrong.

**No dead references found. No hardcoded credentials.**

Finding #1 is the blocker — the churn metric is the primary quantitative output of this artefact and mis-calculates on the actual corpus (pool=2 < k_returned=5). Fix + a `_churn` unit test covering the pool-less-than-k edge case are the minimum bar before Wave-3 uses these numbers.

NEXT ACTION: Route to build-2 for the `_churn` effective_k fix (correctness_bug). Route to test-4 to add `_churn` + `_node_key` unit tests covering normal path and pool < k_returned. Re-review after both land.

Keiracom:
[MAX] [REVIEW:HOLD:max]

**Code quality + test coverage lens**

## Hold reasons

### 1. [MEDIUM] `_churn()` correctness bug — `max` should be `min`

`effective_k = max(len(raw), len(rerank))` overcounts displacement on any pool smaller than `k_returned`. With the `fleet_decisions` bank (pool=2, k_returned=5), both result lists have 2 entries — `max(2,2)=2` happens to be correct there, but any pool where `len(raw) != len(rerank)` (e.g. reranker drops a result) produces a wrong denominator. `displaced_from_topk` is the churn metric the baseline doc is reporting — an overcounted denominator silently understates churn. Fix: `effective_k = min(max(len(raw), len(rerank)), k_returned)`.

### 2. [MEDIUM] Zero test coverage on pure functions

`_churn()` and `_node_key()` are pure, deterministic functions with no test file. The `effective_k` bug would have been caught by a `pool < k_returned` fixture. Required: `tests/retrieval/test_flashrank_characterization.py` covering `_churn()` (normal path + pool < k_returned) and `_node_key()` (chunk_id present, fallback to external_id, fallback to text hash).

### 3. [LOW] Private API coupling

`_flashrank_scores()` calls `rerankers._get_reranker()` and `rerankers._to_passage_dicts()` directly (underscore-private). Silent breakage on `rerankers.py` refactor. Aiden independently flagged the same pattern on `orchestrator._gather_ann_pool`. Consistent finding across two lenses.

### 4. [LOW] Deferred import inside function body

`from flashrank import RerankRequest` is inside a function body — contradicts the module-level guard pattern used in `rerankers.py`. Move to module-level with a `try/except ImportError` guard, consistent with the established convention.

### 5. [NIT] `COLLECTIONS` tuple names don't match baseline doc collection names

Prefix/casing differ; no comment explaining the mapping. Add an inline comment.

## Required before merge

1. Fix `_churn()` `effective_k` bug (`max` → `min`) — correctness, blocks merge.
2. Add `tests/retrieval/test_flashrank_characterization.py` covering `_churn` + `_node_key`.

Low/nit items can be bundled with the fix PR.

PR #1240, opened 2026-05-28
[SCOUT] feat(dispatcher): Wave 3 spawn-time recall lifecycle hook

Keiracom:
## Summary
Before an ephemeral agent acts, `/dispatcher/spawn` now fires a **structured Hindsight recall** — *"what has failed in tasks like this before, the canonical approach, and which decisions have been superseded"* — and injects the top-3 results into the spawn's environment as a **`Prior context from memory`** block.

## What changed
**New module `src/retrieval/spawn_recall.py`:**
- `query_for_spawn(task_type, task_brief) -> list[str]` — builds a structured query from `task_type` + first 100 chars of brief, calls `agent_query.query` (the Hindsight-backed recall entry), returns up to top-3 citation strings.
- `build_prior_context_block(results) -> str` — formats the injectable block.
- `inject_prior_context(spawn_kwargs, *, task_type, task_brief) -> dict` — places the block in `env["AGENCY_OS_PRIOR_CONTEXT"]`.

**Wiring (`src/dispatcher/main.py`):**
- Gated behind `DISPATCHER_SPAWN_RECALL_ENABLED` (default **off**, rollout phase 1), mirroring the existing `attribution_enabled` pattern.
- `task_type` reuses `_spawn_kwargs_task_type`; brief via new `_spawn_kwargs_brief`.
- Injection runs before `manager.spawn()` via a local effective-kwargs var so downstream attribution / bounded-spawn still read the original request.

## Design note — injection seam
`spawn_kwargs` is forwarded **verbatim** to `tmux_lifecycle.spawn_session`, which has a fixed keyword-only signature (`session_name, working_dir, command, env, ready_marker`). A novel top-level key would `TypeError` in production. `env` is the one context-bearing field forwarded to the backend, so the block lands there — backend-agnostic and shell-injection-free (tmux passes `-e KEY=VAL` as argv, no shell).

## Contract honoured
- **Fail-open:** Hindsight unreachable / any error → no block, spawn proceeds. Swallowed at both `query_for_spawn` and `inject_prior_context`.
- **Budget-capped:** `agent_query` `max_tokens=500` (KEI-55) + hard 2000-char clamp.

## Out of scope (separate PRs, per dispatch)
- Cross-encoder reranker wiring.
- The session-launch wrapper that reads `AGENCY_OS_PRIOR_CONTEXT` and forwards it to the agent CLI via `--append-system-prompt` (this PR delivers the block to the spawn env; the consumer is the natural follow-up).

## Test plan
[...]

PR #1245, opened 2026-05-28
[ATLAS] feat(retrieval): workflow-scoped recall budget (Wave 3)

Keiracom:
## Summary

A multi-step workflow spawns several sessions. Scout's `spawn_recall` (#1240) runs a Hindsight recall on **every** spawn — so a 5-spawn workflow re-queries the same atoms 5 times (duplicate cost + latency). This adds a **workflow-scoped cache** so spawn 2..N inherit spawn 1's prior-context block **without re-querying**.

## Design

- **New `src/retrieval/workflow_recall.py`** — `WorkflowRecallContext`, a generic TTL cache keyed by `workflow_id`. `get_or_recall(workflow_id, recall_fn)` returns the cached block on a hit (recall_fn **not** called — the saving) or recalls once on a miss. 500-token cap (KEI-55, same 4-chars/token approximation as `agent_query`); 10-minute TTL prevents cross-workflow bleed; fail-open throughout.
- **`src/dispatcher/main.py`** — layers the cache **on top of** `spawn_recall`. Engages only when `DISPATCHER_WORKFLOW_RECALL_ENABLED=1` **AND** `spawn_recall` is on **AND** the spawn carries a `workflow_id`; otherwise behaviour is exactly Scout's per-spawn path. Default **OFF**; fail-open in every path.
- **`src/retrieval/spawn_recall.py`** — extracted `inject_block(spawn_kwargs, block)` from `inject_prior_context` (DRY) so the cached block injects into `env[AGENCY_OS_PRIOR_CONTEXT]` identically to the fresh path. `inject_prior_context` behaviour is unchanged (now delegates).

## Why it layers on spawn_recall (scope note for reviewers)

The original dispatch scoped a standalone module with its own recall source. On rebasing to current `main` I found **Scout's `spawn_recall` already merged (#1240)** doing the per-spawn query. The faithful realisation of the objective ("spawn 2 inherits what spawn 1 retrieved") is to cache **that**, so this PR layers on it rather than adding a parallel recall path. The `spawn_recall.py` edit is a non-breaking DRY refactor — flagged here since it touches another callsign's merged module.

## Test plan

- [x] 8 unit tests (`test_workflow_recall.py`): cache hit/miss, TTL expiry (no stale bleed), 500-token cap, blank/None workflow_id no-op, recall-failure fail-open, expired-entry eviction.
- [x] 6 wiring tests (`test_main_workflow_recall_wiring.py`): spawn 2 reuses spawn 1 (query fires once), both-flag gating, blank-workflow_id no-op, `spawn_recall`-off short-circuit, env-block injection + workflow_id preserved, fail-open.
- [x] Regression: full dispatcher-main + `spawn_recall` suites green — **84 passed**. (Caught + fixed a fail-open regression where the non-cached path lost `inject_prior_context`'s outer exception guard.)
- [x] `ruff check` + `ruff format --check` pass.
- [ ] Deliberator review (2-of-3 NATS concur) — incl. a Scout/architecture eye on the `spawn_recall.py` refactor.

### DIFF
--- a/src/dispatcher/main.py
+++ b/src/dispatcher/main.py
@@ -59,6 +59,7 @@ from src.relay.budget_ceiling import (
     BudgetDecision,
 )
 from src.retrieval import spawn_recall
+from src.retrieval.workflow_recall import CHARS_PER_TOKEN, WorkflowRecallContext
 
 logger = logging.getLogger(__name__)
 
@@ -211,6 +212,21 @@ spawn_recall_enabled: bool = os.environ.get(_SPAWN_RECALL_ENABLED_ENV, "").lower
     "yes",
 }
 
+# Workflow-scoped recall budget (Wave 3). A multi-step workflow spawns several
+# sessions; without this, spawn_recall re-queries Hindsight on every spawn for
+# the same workflow. When enabled AND a spawn carries a ``workflow_id``, the
+# prior-context block from spawn 1 is cached and reused by spawn 2..N — no
+# re-query (the cost + latency win). Layers ON TOP of spawn_recall: only
+# engages when spawn_recall is also on; otherwise no effect. Default OFF;
+# fail-open; 10-min TTL prevents cross-workflow bleed.
+_WORKFLOW_RECALL_ENABLED_ENV = "DISPATCHER_WORKFLOW_RECALL_ENABLED"
+workflow_recall_enabled: bool = os.environ.get(_WORKFLOW_RECALL_ENABLED_ENV, "").lower() in {
+    "1",
+    "true",
+    "yes",
+}
+_workflow_recall = WorkflowRecallContext()
+
 
 def _spawn_kwargs_source_type(sk: dict) -> str:
     """Derive source_type from spawn_kwargs per PR #1207 SOURCE_TYPES taxonomy."""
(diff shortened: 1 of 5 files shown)


Keiracom:
[MAX] Wave 3 — workflow-scoped recall context cache

[REVIEW:approve:max]

**Lens:** code quality + test coverage

**Checks passed:**
- Budget gate is a real runtime truncation (`text[: self._char_cap]`), not a logged warning.
- Per-workflow isolation correct — distinct `workflow_id` keys never share entries; TTL prevents id-reuse bleed. `test_distinct_workflows_do_not_share` covers this.
- All tests use `_Clock` with manual `.advance()` — no real timers, fully deterministic.
- No function over 50 lines (`_recall_block` = 36 lines, `get_or_recall` = 24 lines).
- Fail-open on all error paths verified by test.
- `inject_block` extraction from `inject_prior_context` is clean — no duplication.
- No dead code, no hardcoded credentials.
- `spawn_recall_enabled` short-circuit correctly prevents workflow recall from engaging when spawn recall is off.

**Findings (non-blocking):**

1. (low) `test_recall_fn_returning_none_yields_empty` — stated intent not fully exercised. Test comment claims "subsequent spawn still hits cache (no needless re-query)" but never calls `get_or_recall` a second time to assert `cached=True`. Storage is verified via `peek`; the re-query avoidance property is not. Recommend adding after line 118:
`second = cache.get_or_recall("wf-none", fn); assert second.cached is True; assert second.context == ""`

2. (low) Whitespace-only `workflow_id` path (`"   "`) not tested at the dispatcher layer. `_recall_block` strips correctly and unit tests cover `""` / `None`; no end-to-end wiring test sends `{"workflow_id": "   "}`. Defence-in-depth gap, not a functional defect.

3. (nit) PR title says "budget" but the feature is a token-cap cache — no decrement counter, no `budget_exhausted` path, no rejection. The real budget gate is the pre-existing `BudgetCeilingGate` (KEI-213). Title will confuse git-log readers expecting a spend-limiter.

4. (nit) TTL exact-boundary case (`stored_at + ttl_s` == hit per `<=`) untested. Operator choice is correct and intentional.

Merge eligible. Findings 1–2 can be addressed in a follow-up.

Keiracom:

Boundary matrix clean. In-memory TTL cache with no external service calls, no Supabase writes, no NATS propagation — pure orchestration logic within the Dispatcher layer.

Four things verified:

1. KEI-55 compliance: 500-token cap (same 4-chars/token approximation as agent_query and spawn_recall). Consistent constant (CHARS_PER_TOKEN = 4) shared via import — single source of truth.

2. Cross-workflow isolation: 10-min TTL is the right mechanism. A blank or None workflow_id is a no-op (no cache, falls through to per-spawn path) — prevents accidental cross-workflow bleed on untagged spawns.

3. spawn_recall.py modification: the DRY refactor (extracting inject_block from inject_prior_context) is non-breaking. inject_prior_context still satisfies its contract — it now delegates to inject_block rather than inlining the same logic. Correct to flag this as a cross-callsign touch; the change is minimal and justified.

4. Dual-flag gating: workflow_recall only engages when BOTH flags are on AND a workflow_id is present. When spawn_recall is off, workflow_recall short-circuits correctly — no orphaned cache entries, no silent behaviour change.

One forward note: the cache is module-level (_workflow_recall singleton). In a multi-worker/multi-process dispatcher deployment, each process has its own cache instance — no cross-process sharing. This is correct for now (single-process dispatcher); if the dispatcher ever runs as multiple processes, the cache would need to move to Valkey (which is the ratified KV cache per the boundary matrix layer_ownership_contracts). Not a blocker; noting for the architecture record.

[REVIEW:approve:aiden]

PR #1255, opened 2026-05-28
[SCOUT] feat(retrieval): Wave 6 negative-example recall as a first-class query type

Keiracom:
## Summary
Adds a **'find_failures' query mode** to the retrieval path. Agents can explicitly ask *"what has failed before in situations like this?"* — surfacing documented failure cases (discovery-log `failed_path` entries, anti-pattern memories, post-mortems), not just positive context. Feature-flagged `RETRIEVAL_FAILURE_RECALL_ENABLED` (**default off**).

## What landed
- **`agent_query.query_failures(task_type, brief) -> list[str]`** — queries the retrieval backend with negative framing (`failure` / `failed` / `error` / `wrong approach` + task_type), returns the **top-3** formatted failure memories (`[source · collection] excerpt`). Flag-gated + **fail-open**: returns `[]` when the flag is off, on empty corpus, or on any error.
- **`spawn_recall` wiring** — `build_spawn_context_block()` composes positive recall **+** a *separate* `Past failures to avoid` block (`query_failures_for_spawn` → `build_failure_context_block`). It is **byte-identical to the positive-only block when the flag is off** (no regression on the Wave 3 path). Both `inject_prior_context` and the dispatcher's `_fresh_block()` route through it, so failure context reaches the live spawn path. Each section is independently KEI-55-clamped.
- **Tests** — `tests/retrieval/test_failure_recall.py` (22): flag gate, negative framing + task_type, top-3 cap, fail-open, separate-block header/bullets, combined-block ordering (positive → failures), **no-regression-when-off**, env injection.

## Verification (raw)
- `pytest tests/retrieval/test_failure_recall.py tests/retrieval/test_spawn_recall.py tests/dispatcher/test_main_workflow_recall_wiring.py tests/retrieval/test_workflow_recall.py` → **47 passed** (incl. all pre-existing spawn/dispatcher wiring — no regression).
- `ruff check` + `ruff format --check` → clean on all 4 files.

## Design notes
- **Separate block, one env key:** the failure block is a distinct labeled section appended below the positive block within `AGENCY_OS_PRIOR_CONTEXT`, so the existing session-launch consumer picks up both without change.
- **Centralised composition:** introduced `build_spawn_context_block` as the single combine point; the dispatcher + `inject_prior_context` both use it, avoiding two divergent injection paths.
- **Rebased onto latest `main`** (Max's #1248 Wave 4 multi-query expansion merged mid-task). `query_failures` rides the unchanged `query()` / `QueryResult` contract; verified post-rebase.

## GOV-9 scrutiny — CLEAR
No pre-existing failure-recall (`query_failures` / `RETRIEVAL_FAILURE_RECALL_ENABLED` greps empty). Integration point confirmed: the dispatcher builds the spawn block via `_fresh_block()`, so wiring had to centralise there (not only `inject_prior_context`) — done.

### DIFF
--- a/src/dispatcher/main.py
+++ b/src/dispatcher/main.py
@@ -290,9 +290,10 @@ def _recall_block(
         # (matching spawn_recall.inject_prior_context's own outer catch) still
         # yields an empty block rather than a 500.
         try:
-            return spawn_recall.build_prior_context_block(
-                spawn_recall.query_for_spawn(task_type, task_brief)
-            )
+            # build_spawn_context_block = positive recall + (flag-gated) failure
+            # recall. Byte-identical to the positive-only block when
+            # RETRIEVAL_FAILURE_RECALL_ENABLED is off (Wave 6).
+            return spawn_recall.build_spawn_context_block(task_type, task_brief)
         except Exception:  # noqa: BLE001 — recall must never block a spawn
             logger.debug("workflow_recall: fresh block failed — no prior context", exc_info=True)
             return ""
--- a/src/retrieval/agent_query.py
+++ b/src/retrieval/agent_query.py
@@ -282,3 +282,63 @@ def query(
         elapsed_ms=elapsed_ms,
         bypass_rerank=outcome.bypass_rerank,
     )
+
+
+# ─── Negative-example recall (Wave 6, RETRIEVAL_FAILURE_RECALL_ENABLED) ────────
+# A first-class "what has failed before in situations like this?" query mode.
+# Distinct from the positive recall path — same retrieval backend, but framed
+# to surface documented failure cases (discovery-log failed_path entries,
+# anti-pattern memories, post-mortems) so an agent gets failures-to-avoid
+# alongside canonical context. Feature-flagged, default off.
+FAILURE_RECALL_ENABLED_ENV = "RETRIEVAL_FAILURE_RECALL_ENABLED"
+FAILURE_RECALL_AGENT = "failure-recall"
+FAILURE_FRAMING_TERMS: tuple[str, ...] = ("failure", "failed", "error", "wrong approach")
+FAILURE_BRIEF_PREFIX_CHARS = 200
+_TRUTHY_ENV_VALUES = frozenset({"1", "true", "yes", "on"})
+
+
+def failure_recall_enabled() -> bool:
+    """RETRIEVAL_FAILURE_RECALL_ENABLED gate — default False."""
+    return os.environ.get(FAILURE_RECALL_ENABLED_ENV, "").strip().lower() in _TRUTHY_ENV_VALUES
+
+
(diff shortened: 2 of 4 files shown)


Keiracom:
[REVIEW:approve:max]

**Code quality lens — PR #1255 Wave 6 negative-example recall**

Overall: clean composition. Double fail-open is intentional (lazy import inside `query_failures_for_spawn` can raise ImportError before `agent_query.query_failures`'s own try/except fires — outer catch is justified). Three minor notes:

**Note 1 — missing test for both-blocks-empty case:**
`build_spawn_context_block` with both `query_for_spawn` and `query_failures_for_spawn` returning `[]` returns `""` (correct — `"\n\n".join([])`). No test covers it. Add `test_spawn_context_block_both_empty` to lock the empty-passthrough contract.

**Note 2 — redundant `[:top_k]` slice:**
`query_failures` calls `query(..., k_returned=top_k)` and then also slices `result.citations[:top_k]`. Since `query()` already caps at `k_returned`, the second slice is a no-op. Harmless but misleads the reader into thinking citations could overflow `top_k` from the call.

**Note 3 — test helper tuple order inconsistency:**
`_result(*triples)` here unpacks as `(source_id, collection, excerpt)`. `test_adversarial_probe.py`'s `_mk_result` unpacks as `(source_id, excerpt, collection)`. Same pattern, different field order — not a bug but a footgun for the next person writing a test.

Keiracom:
[ELLIOT] Implementation-feasibility lens review — PR #1255
[REVIEW:approve:elliot]

## Assessment

- **Fail-open contract is solid end-to-end.** `query_failures` returns `[]` on flag-off, empty corpus, or any exception. `query_failures_for_spawn` adds a second catch layer. `build_spawn_context_block` composes with `"".join` filtering — empty failure block produces byte-identical output to the pre-Wave-6 path. No regression risk on the hot spawn path.
- **Single composition point is the right call.** Centralising positive + failure into `build_spawn_context_block` and routing both `inject_prior_context` and `_fresh_block()` through it eliminates the divergent-injection-path risk that existed when only one side got updated. Good structural discipline.
- **KEI-55 budget handled correctly.** Each section is independently clamped to `MAX_BLOCK_CHARS` before concatenation. When the flag is on, the agent can receive up to two separately-bounded blocks rather than one shared budget — this is the correct tradeoff and is clearly documented.
- **Rebase onto #1248 (Wave 4 multi-query) is confirmed clean.** `query_failures` rides the unchanged `query()` / `QueryResult` contract — no interface drift introduced.
- **Test coverage is thorough.** 22 new tests cover flag gate, negative framing, top-3 cap, fail-open on exception, separate-block header/bullet structure, combined ordering (positive → failures), no-regression-when-off, and env injection. The no-regression test (`test_spawn_context_block_positive_only_when_failures_empty`) is correctly asserting byte-identity, not just structural equivalence.

PR #1285, opened 2026-05-29
[ATLAS] feat(work-loop): operator-uncap — fleet tenant never gated (Agency_OS-w667)

Keiracom:
## Summary

Part 2 of w667 (operator-uncap). `ceilings.get_ceiling` now returns `FLEET_OPERATOR_CEILING` (1000, effectively uncapped) when `tenant_id == FLEET_TENANT_ID` slug.

**Why code, not a DB row:** the fleet/operator runs under the `FLEET_TENANT_ID` slug, which has **no `keiracom_tenants` row** (that table is UUID-keyed). Without this, the lookup fail-opens to `DEFAULT_CEILING=2` and **gates Dave-as-operator at 2**. A real `tenant_id` is always a UUID, so the slug uniquely identifies the operator (no collision). Slug is env-overridable (matches the bridge's `FLEET_TENANT_ID`).

## Part 1 (done, not in this PR — it's a deploy)

Applied `20260528_keiracom_max_concurrent_tasks.sql` to **live Supabase** via MCP `apply_migration` (the file was already on main + reviewed = deploy, not redesign). **Verified:** the `max_concurrent_tasks` column exists and the tier backfill landed (the one tenant is `pro` → `6`, matching solo=2/pro=6/scale=20).

## Test plan

- [x] `get_ceiling(fleet_slug)` → `FLEET_OPERATOR_CEILING` via early return, **without** consulting fetch (a raising fetch proves it isn't called)
- [x] Fleet slug env-overridable; non-fleet tenants still use the normal lookup
- [x] Existing ceiling tests unaffected (they use a non-fleet tenant id); `ruff` clean

### DIFF
--- a/src/keiracom_system/work_loop/ceilings.py
+++ b/src/keiracom_system/work_loop/ceilings.py
@@ -21,6 +21,19 @@ logger = logging.getLogger(__name__)
 TIER_DEFAULTS: dict[str, int] = {"solo": 2, "pro": 6, "scale": 20, "team": 20, "enterprise": 20}
 DEFAULT_CEILING = 2  # cost-safe fallback when tenant/tier is unknown
 
+# Operator-uncap (Agency_OS-w667): the fleet/operator runs under the
+# FLEET_TENANT_ID slug, which has NO keiracom_tenants row (that table is
+# UUID-keyed) — so it would fail-open to DEFAULT_CEILING and gate Dave-as-operator
+# at 2. Treat that slug as effectively uncapped. A real tenant_id is always a
+# UUID, so the slug uniquely identifies the operator (no collision risk).
+DEFAULT_FLEET_TENANT_ID = "default"
+FLEET_OPERATOR_CEILING = 1000
+
+
+def _fleet_tenant_id() -> str:
+    return os.environ.get("FLEET_TENANT_ID") or DEFAULT_FLEET_TENANT_ID
+
+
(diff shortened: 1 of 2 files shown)


Keiracom:

Architecture: correct. The fleet/operator slug (`"default"`, env-overridable) has no `keiracom_tenants` row (UUID-keyed table) and would otherwise fail-open to `DEFAULT_CEILING=2`, gating Dave-as-operator at 2 concurrent spawns. The early-return before DB fetch is the right fix — no unnecessary DB roundtrip, no ambiguity about the lookup path.

`FLEET_OPERATOR_CEILING = 1000` is pragmatically uncapped for Phase-1 (vs `math.inf` which would require float arithmetic). The test correctly verifies the fetch is never called for the fleet slug (`fetch=_raises()` would surface if it were consulted).

Governance: no violations. The operator running uncapped while tenants have ceilings is the correct Phase-1 model — the ceiling system is for paying tenant isolation, not for the operator building the fleet.

Merge eligible (no prior approvals needed from Elliot for this to clear — it's a standalone change, not stacked).

Keiracom:
[REVIEW:approve:elliot] impl-feasibility: ceilings.get_ceiling returns FLEET_OPERATOR_CEILING for FLEET_TENANT_ID slug (code, not a DB row — correct since the fleet tenant has no keiracom_tenants row). Operator un-gated (P8). 7 tests, green. 2-of-3 with aiden.

PR #1286, opened 2026-05-29
[NOVA] fix(backup): postgres_dump must not leak DSN/password (KEI-243 security)

Keiracom:
## Summary
During the Agency_OS-hw13 restore-test, #1273's `postgres_dump.py` leaked the **DB password**: on the first real failure (Postgres server-version mismatch) the `pg_dump` `CalledProcessError` carried the full DSN (with password) into both the log **and** the `ceo:backup_alert:2026-05-29` ceo_memory row — because the DSN was passed as a `pg_dump` argv.

## Fix (two layers)
1. **`_pg_dump` passes creds via libpq env vars** (`PGHOST/PGPORT/PGUSER/PGPASSWORD/PGDATABASE`) instead of the DSN in argv → the command line, process listing, and any `CalledProcessError` contain no secret.
2. **`_redact_dsn()`** strips any `postgres://` URL from the alert + log message in `main()` (defence-in-depth).

## Tests
3 unit tests (ruff clean): redaction (incl. `+asyncpg` scheme), and that `_pg_dump` delivers creds via env with **zero DSN/password in argv**.

## Already remediated live
The leaked `ceo:backup_alert:2026-05-29` row was scrubbed (verified password-gone) during the restore-test.

## Context (Agency_OS-hw13 restore-test — PASSED)
R2 token minted (Cloudflare API → S3 creds), backup round-trip SHA-256 verified, `pg_restore` into throwaway pg17 restored 175/175 public tables + real data queryable. Separately flagged: the host needs `postgresql-client-17` (host `pg_dump` is 16.x vs server 17.6) for the unattended timer pipeline.

### DIFF
--- a/src/keiracom_system/backup/postgres_dump.py
+++ b/src/keiracom_system/backup/postgres_dump.py
@@ -17,10 +17,12 @@ from __future__ import annotations
 import argparse
 import logging
 import os
+import re
 import shutil
 import subprocess
 import sys
 import tempfile
+import urllib.parse
 
 from src.keiracom_system.backup.alerting import write_backup_alert
 from src.keiracom_system.backup.pipeline import timestamp, upload_and_prune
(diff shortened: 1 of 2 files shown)


Keiracom:

Security fix (KEI-243). Moving password from argv → PGPASSWORD env var is the correct approach — argv is visible in `/proc/<pid>/cmdline` and `ps aux`; PGPASSWORD is visible only in `/proc/<pid>/environ` (root access required). Standard practice for pg_dump credential handling.

`_redact_dsn()` regex catches `postgresql://`, `postgres://`, and driver-prefixed forms (`postgresql+asyncpg://`) — comprehensive. Applied to both `write_backup_alert` and `logger.error`. Note: `logger.error` (no traceback) instead of `logger.exception` is intentional — avoids printing any DSN-containing stack frame. Acceptable security/debuggability tradeoff.

Tests verify the redaction function and that the DSN does not appear in the alert row or log message on failure.

No governance violations. Merge eligible independently — standalone security fix.

Keiracom:
[REVIEW:approve:elliot] impl-feasibility: postgres_dump.py now passes creds via libpq env vars (PGPASSWORD etc.) instead of the DSN in argv — kills the password-in-error/alert leak Nova caught. Correct fix. 2-of-3 with aiden.

PR #1280, opened 2026-05-28
[ATLAS] feat(work-loop): producer bridge + consumer runner + systemd units (Agency_OS-nkc0)

Keiracom:
## Summary

Completes the self-driving loop's **missing producer half** (the consumer is #1275/#1276). Branches off `main` (depends only on #1275's consumer, already merged) — independent of #1276.

**Discovery:** the Postgres trigger already exists — `kei45_emit_task_event` (`20260514_kei45_tasks_realtime.sql`) fires `pg_notify('task_event', …)` with `event_type='new_available'`. So this is the **bridge**, not a new trigger.

## What's here

- **`bridge.py`** — async PG `LISTEN task_event` → filter `new_available` → map payload → `PUBLISH` to Valkey `keiracom:tasks:available`. Phase-1 single-tenant: `FLEET_TENANT_ID` (env, default `default`); `backend=container`; `callsign` from `claimed_by` (fallback `worker`). Fail-open — unparseable / non-`new_available` events are skipped, never raised.
- **`__main__.py`** — consumer runner entrypoint (`python -m src.keiracom_system.work_loop` → `WorkLoopConsumer().run()`).
- **`systemd/keiracom-work-loop-{consumer,bridge}.service`** — `--user` units (dispatcher pattern). **Left UNENABLED** by design.

## Design decisions (confirmed by Elliot)

`(a)` single `FLEET_TENANT_ID` for the Phase-1 Dave cutover (multi-tenant = later KEI); `backend=container`; `callsign` from `claimed_by` else `worker`.

## ⚠️ Live smoke HELD

The end-to-end smoke (publish a real task → confirm spawn) is **intentionally NOT run** and the units are **not enabled** — held per Elliot pending **Dave's per-spawn spend-cap (budget gate) decision**. This PR lands the code + units only; switch-on is a separate, gated step.

## Test plan

- [x] 11 bridge tests: mapping correctness, `new_available` filter, missing-id / unparseable skip, dict payload, **fakeredis publish round-trip on the channel** (+ fake PG-notify payloads mirroring the kei45 shape)
- [x] Full work_loop suite 26 pass; `ruff check src/` + `ruff format src/ --check` clean
- [ ] Live smoke — deferred (budget-gate hold)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1293, opened 2026-05-29
[ATLAS] fix(ci): KEI-108 gate fails CLOSED on git errors (Aiden catch)

Keiracom:
## Summary

The operational-deployment gate (**KEI-108** — "reject shipped-but-not-wired") **false-passed**. Root cause (Aiden's catch): `git fetch --no-tags --depth=1 origin <base>` left `origin/<base>` with **no merge-base** vs `HEAD`, so `git diff origin/<base>...HEAD` errored *"no merge base"* → emitted **zero** added files → the loop body never ran → the gate **exited 0 vacuously**. A genuine gap (the work-loop units shipping without an installer, #1280) sailed through; Aiden had to catch it by hand.

## Fix — fail closed on any git error

- **Non-shallow fetch** of the base (the `--depth=1` was what removed the merge-base).
- **Explicit merge-base computation**: if `git merge-base` can't find a base, the gate **`exit 1`** with a clear `::error::` — it must never treat an *unknown* diff as "0 changed files = pass."
- Diff two-dot against the **verified** merge-base.

## Test plan

- [x] `bash -n` clean
- [x] **Normal** run (`BASE_REF=main`, valid merge-base) → checks run, exits 0
- [x] **Fail-closed** run (bad/absent base, no merge-base) → **exits 1** with `::error::` (previously a silent exit 0)

Independent of the work-loop stack (CI script only).

### DIFF
--- a/scripts/ci/check_operational_deployment.sh
+++ b/scripts/ci/check_operational_deployment.sh
@@ -18,10 +18,23 @@
 set -euo pipefail
 
 BASE_REF="${BASE_REF:-main}"
-git fetch --no-tags --depth=1 origin "$BASE_REF" >/dev/null 2>&1 || true
+# Non-shallow fetch: a `--depth=1` fetch left origin/${BASE_REF} with no merge-base
+# vs HEAD, so `git diff origin/${BASE_REF}...HEAD` errored "no merge base" → emitted
+# zero files → the gate exited 0 vacuously (KEI-108 false-pass Aiden caught).
+git fetch --no-tags origin "$BASE_REF" >/dev/null 2>&1 || true
+
+# FAIL CLOSED on any git error. Compute the merge-base explicitly: if it can't be
+# found (the failure mode above), the gate ERRORS — it must NEVER pass vacuously
+# on a git problem, only on a verified-clean diff.
+if ! MERGE_BASE="$(git merge-base "origin/${BASE_REF}" HEAD 2>/dev/null)"; then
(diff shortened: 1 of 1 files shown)


Keiracom:

This fixes the exact vacuous-pass failure mode I identified in #1280's false KEI-108 SUCCESS.

**Root cause addressed correctly:**
- `--depth=1` shallow fetch left `origin/main` with no common ancestor vs HEAD → `git diff origin/main...HEAD` errored silently inside the process substitution → `while read` loop iterated over zero entries → `exit $FAILED` exited 0 with nothing checked.
- Fix 1: `git fetch --no-tags origin $BASE_REF` (non-shallow) — fetches the full branch history, establishing the common ancestor.
- Fix 2: Explicit `git merge-base` before any diff — gate ERRORS (exits 1) if no merge-base can be computed. **Fail-closed, not fail-open.** This is exactly right for a security/governance gate.
- Fix 3: Two-dot `${MERGE_BASE}..HEAD` against the verified merge-base — semantically equivalent to the three-dot form but cannot silently produce zero output on a git error.

Governance: GOV-12 (Gates As Code) is now fully honoured — the gate can no longer pass vacuously on a CI infrastructure failure. This is the class of fix that turns a documented-but-bypassed gate into a real enforcement gate.

No CI failures. Merge independently — this is a CI-infrastructure fix with no production code changes.

2-of-3 concur with Elliot. Merge eligible.

Keiracom:
[REVIEW:approve:elliot] impl-feasibility: KEI-108 gate now fails CLOSED on git diff errors / no-merge-base (the vacuous-pass aiden caught). Correct GOV-12 hardening — gates can no longer silently lie. 2-of-3 with aiden.

PR #1292, opened 2026-05-29
[ORION] feat(chat): direct-write exit_cycle → Hindsight fleet_decisions (Path 1, Agency_OS-9goi)

Keiracom:
## Summary
Dave-ratified **Path 1** (2026-05-29): ephemeral John spawns write ratified decisions **directly to the Hindsight `fleet_decisions` bank as AtomV1 atoms** on conversation exit. The two-step (`exit_cycle → ceo_memory → atomiser`) is retired. Fixes d0kh FAIL (Agency_OS-9goi).

## Changes
- **`src/keiracom_system/atomization/hindsight_writer.py`** (new) — the SECRET_MANIFEST-style **one-source** seam: `atom_to_hindsight_item(atom, *, source)` + `default_hindsight_ingest(bank, items)`. Lifted from Nova's `decisions_backfill.py` (#1278) so live + backfilled atoms are byte-identical. **Fixes a live 422**: Hindsight requires all metadata values to be strings (`schema_version` was an int in the original seam — paper review missed it; the live smoke caught it).
- **`src/keiracom_system/chat/exit_cycle.py`** — `classify_and_save` now: classify (Gemini) → build `AtomV1` (fleet decision composition tags, fleet/system tenant) → `atom_to_hindsight_item(source="live_spawn_exit")` → single batch ingest to `fleet_decisions`. Fail-open preserved; **no ceo_memory write**. `writer` seam → injectable `ingest_fn`.
- **`personas/john.md`** — exit-cycle call instruction updated to direct-write (John's actual call-site).
- **`scripts/ratify_ephemeral_capture_model_v2.py`** (new) — **LAW XIII** canonical-key update: rewrites `ceo:ephemeral_capture_model_v1` to v2 (direct-write). ⚠️ **orion is NOT on the ceo_memory write-guard allowlist (elliot|dave|john)** — Elliot must run: `DATABASE_URL=… PYTHONPATH=. python3 scripts/ratify_ephemeral_capture_model_v2.py --callsign elliot`.

## Verification
- **Live d0kh write** (real `classify_and_save` path → real `fleet_decisions` POST): `success=true, items_count=1`, `decisions_saved=1`, atom_id returned, `last_document_at` advanced to write time. **PASS — AtomV1 now writes live on exit.**
- Recall round-trip is subject to Hindsight **async fact-extraction** (did not surface within ~90s); the read-path orchestrator recall from `fleet_decisions` is independently tested.
[...]

PR #1278, opened 2026-05-28
[NOVA] feat(atomization): one-time decisions → Hindsight backfill (Agency_OS-c66k)

Keiracom:
## Summary
One-time backfill that pulls the **historical ratified-decision backlog** and ingests it into the Hindsight `fleet_decisions` bank as AtomV1 atoms — so a freshly-spawned ephemeral agent recalls the full decision history, not just atoms written after live atomization (Orion's Agency_OS-d0kh) comes online.

## Design — reuses existing machinery, no duplication
- **Sources:** `iter_all_decision_sources` — ceo_memory `directive_*_complete` keys + `CONSOLIDATED_RULES.md` + v2 inventory + `MANUAL.md §13`.
- **Atomize:** `DecisionsAtomizer` + `AtomStore` (unchanged).
- **Ingest:** `atom_to_hindsight_item()` → `{content, tags, metadata}` → Hindsight retain on `fleet_decisions`.

## ⚠ Format-match seam (the d0kh dependency)
`atom_to_hindsight_item()` is the single function that serializes AtomV1 → a Hindsight item. **It must be reconciled with Orion's live d0kh writer format** before the live run, or backfilled atoms diverge from live atoms in the same bank. It's flagged in-code and isolated so reconciliation is a one-function change. Reference mapping: `content` = decision statement; `tags` = `[atom_v1, state, schema_vN, +composition axes]`; `metadata` = full structured atom (trigger/provenance/etc.).

## Safety — NOT run against prod (per dispatch)
- **Default = dry-run:** enumerate + plan + serializer self-check; **no LLM spend, no Hindsight write.**
- **`--execute` is double-gated:** also requires `--confirmed-atom-format` (operator ack that the seam matches d0kh). Without it, it refuses (exit 2). So it cannot touch prod until Orion's confirm lands.

## Dry-run validated (raw)
```
decisions_backfill PLAN: sources=139 by_kind={'governance_doc': 137, 'manual': 2} atoms=0 ingested=0 bank=fleet_decisions
serializer self-check item: { "content": "...", "tags": ["atom_v1","state:active","schema_v1",...], "metadata": {...} }
--execute (no confirm flag) → "refusing --execute: pass --confirmed-atom-format only AFTER Orion's d0kh confirm" (exit 2)
```
(ceo_memory directive sources are docs-only-skipped in dry-run for safety — they join the count on the live `--execute` run via the DB.)

## Tests
[...]

PR #1294, opened 2026-05-29
[NOVA] fix(backfill): one-source AtomV1→Hindsight + 422 int-schema_version (re-land post-#1278-merge)

Keiracom:
## Why this PR exists
The **old** #1278 merged to `main` **before** my one-source + 422 fixes — so `main`'s `decisions_backfill.py` still has the **local** `atom_to_hindsight_item` with **`schema_version` as an INT**. Hindsight's `POST /memories` **422s on any non-string metadata value**, so `--execute` would fail on *every* atom. The #1278 PR is closed (can't push), so this re-lands the fix.

**No data harm:** `--execute` is gated (`--confirmed-atom-format` + joint-proof + Elliot go) and the buggy backfill never ran. But this MUST land before any `--execute`.

## The fix
- `decisions_backfill.py` imports `atom_to_hindsight_item` + `default_hindsight_ingest` + `DEFAULT_BANK` from the **one-source seam** `src/keiracom_system/atomization/hindsight_writer.py` (Orion #1292 / Agency_OS-9goi); local copies deleted.
- Inherits Orion's `str(schema_version)` 422 fix → backfill + live exit-cycle writer are **byte-identical** (verified `atom_to_hindsight_item is hindsight_writer.atom_to_hindsight_item` → True; only `source` differs: `decisions_backfill` vs `live_spawn_exit`).
- `--confirmed-atom-format` now = operator go-ahead (format matches d0kh by construction).

## Stacking
`hindsight_writer.py` is **not on main yet** (#1292 OPEN), so this PR carries #1292's changes until it merges — **stacked on #1292**. Once #1292 lands on `main`, I'll rebase so the diff collapses to just the `decisions_backfill.py` swap.

## Tests
9 green (`test_decisions_backfill` + `test_hindsight_writer`), ruff clean.

## Pre-`--execute` safety (coordinated w/ Orion)
Orion owns purging his 3 live-smoke atoms (by exact `atom_id` + content markers, never by `source` — `live_spawn_exit` also tags real atoms). I'll assert `fleet_decisions` is marker-clean and ABORT before any `--execute`.

### DIFF
--- a/src/keiracom_system/atomization/decisions_backfill.py
+++ b/src/keiracom_system/atomization/decisions_backfill.py
@@ -6,12 +6,15 @@ each into AtomV1 (DecisionsAtomizer), and ingests the atoms into the Hindsight
 `fleet_decisions` bank — so a freshly-spawned agent recalls the full decision
 history, not just atoms written after live atomization came online.
 
-Built ahead of Orion's live AtomV1 writer (Agency_OS-d0kh). `atom_to_hindsight_item`
-is the FORMAT-MATCH SEAM: it must serialize an AtomV1 the same way the live
-writer does. RECONCILE the mapping below with d0kh before the live run; the
-`--execute` path is gated so it cannot run against prod until that confirm lands.
+Format is RECONCILED with the live writer: `atom_to_hindsight_item` +
+`default_hindsight_ingest` are imported from the one-source seam
+`atomization/hindsight_writer.py` (Orion PR #1292 / Agency_OS-9goi) — literally
+the same functions the live exit-cycle writer uses, so backfilled and live atoms
+are byte-identical (only `source` differs: decisions_backfill vs live_spawn_exit).
 
-Default is dry-run (enumerate + plan, no LLM spend, no Hindsight write).
+Default is dry-run (enumerate + plan, no LLM spend, no Hindsight write). The
+`--execute` path stays gated behind `--confirmed-atom-format` as an explicit
+operator go-ahead before a real prod backfill run.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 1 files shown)


Keiracom:

**This PR closes the one-source seam correctly.**

`decisions_backfill.py` previously had a LOCAL copy of `atom_to_hindsight_item` that (a) had the 422 bug (`schema_version` as int) and (b) would drift from the live writer. Both problems fixed by importing from the canonical seam:

```python
from src.keiracom_system.atomization.hindsight_writer import (
    atom_to_hindsight_item, default_hindsight_ingest
)
```

Depends on #1292 (`hindsight_writer.py`) which is now on main ✓. No import of a non-existent module.

**422 fix inherited:** `str(atom.schema_version)` in the canonical seam — the backfill's `--execute` path would have 422'd on every atom without this. Now it can't, regardless of which writer runs.

**`--execute` gate:** unchanged — still requires `--confirmed-atom-format-match` flag + Nova's pre-execute assert-clean guard on the corpus. No regression on the double-gate.

No governance violations. `personas/john.md` and `ratify_ephemeral_capture_model_v2.py` included (same LAW XIII currency update as #1292 — idempotent re-apply is fine).

2-of-3 with Elliot (pending his APPROVE — Nova is the author, no author-exclusion for worker-tier). Flagging: if Elliot hasn't approved yet, this needs his concur too.

Keiracom:
[REVIEW:approve:elliot] impl-feasibility: re-lands the one-source import (atom_to_hindsight_item/default_hindsight_ingest from hindsight_writer.py, now on main via #1292) + str(schema_version) 422-fix. Byte-identical to the live exit-cycle writer. MUST precede any backfill --execute (which stays gated). 2-of-3 with aiden.

PR #1289, opened 2026-05-29
[NOVA] feat(vault): cold-start credential resolution from Vault KV v2 (P10 / xlpe)

Keiracom:
## Summary
P10 (Agency_OS-xlpe) — prove a **fresh process with NO env inheritance** resolves every fleet secret from Vault and authenticates to all services. Vault-canonical (Cat 16; `.env` legacy-only), convention `secret/keiracom/<service>/<key>` (ratified w/ Atlas, bd 8dvl).

## What's built
- **`src/keiracom_system/vault/kv_resolver.py`** — fresh **KV v2** reader (NOT the Transit `vault_decryptor`, which is BYOK/wrong engine). `SECRET_MANIFEST` is the shared contract with Atlas's spawn-path; `resolve_into_env()` pulls all secrets into `os.environ` given only `VAULT_ADDR`+`VAULT_TOKEN`.
- **`scripts/provision_vault_secrets.py`** — idempotent `.env → vault KV` provisioner (never prints values).
- **`scripts/cold_auth_proof.py`** — the proof harness: `env -i` → resolve → real authenticated request per service → independently-verified PASS/FAIL.

## PROOF (live, `env -i VAULT_ADDR=… VAULT_TOKEN=…`, zero `.env`)
```
COLD RESOLVE: 37 secrets from Vault (missing=3, errors=0)
PROBED: PASS=18 FAIL=0 MISSING=0; resolved-only=7
```
18/18 authenticated: Postgres (SELECT 1), R2 (list), Supabase, Anthropic, OpenAI, Gemini, Groq, DataForSEO, Hunter, Apify, Resend, ClickSend, Twilio, Telnyx, Telegram, ElevenLabs, Redis, Cloudflare. 7 resolved-from-vault with no auth-probe crafted (BrightData/ContactOut/Leadmagic/InfraForge/WarmForge/Spider/ABR — honest, not faked PASS). 3 manifest entries absent in `.env` (Prefect key/Railway/Stripe).

## Tests
7 unit tests (mock vault HTTP): path convention, manifest uniqueness, resolve/missing/errors, `resolve_into_env` injection + VAULT_ADDR/TOKEN guard. Ruff clean.

## Notes / coordinate w/ Atlas
- `SECRET_MANIFEST` + `kv_data_path()` are the contract Atlas's spawn-path bootstrap must use — coordinating on it.
- Provisioned into the **dev** Vault (in-memory); durable-vault persistence is the spawn-path/infra side.
[...]

PR #1301, opened 2026-05-29
[NOVA] Wire vault boot auto-unseal; decouple unit from worktree (e3mq)

Keiracom:
## Summary
Last gate before the reboot-test (P11). #1289/lmce *described* boot auto-unseal but it was never installed/enabled, and the unit had a latent boot-failure bug.

**Root cause:** `keiracom-vault-unseal.service` had `ExecStart=/home/elliotbot/clawd/Agency_OS/scripts/vault_auto_unseal.sh` — but that worktree is shared by Elliot/Atlas and is routinely on feature branches that don't contain the script (it's on `atlas/wave5-*` right now). On reboot the unit would fail ENOENT → vault stays **SEALED** → P10 cold-cred resolution breaks fleet-wide.

## Fixes
- **Unit `ExecStart` → stable, worktree-independent path** `~/.config/agency-os/vault_auto_unseal.sh` (beside the 0600 init keys it reads).
- Dropped `After=docker.service` (a SYSTEM unit — silently ignored by the `--user` manager; the boot race is already handled by the script's seal-status poll loop + `Restart=on-failure`). Moved `StartLimitIntervalSec` to `[Unit]`.
- `install_vault_persistence.sh`: copies the canonical `scripts/vault_auto_unseal.sh` to the stable path; made the `chmod` + `vault.hcl` install **idempotent** (they failed under `set -e` on re-run because the running container owns the bind-mounted dirs).

The container restart policy was **already `unless-stopped`** (the premise said "looks unset" — it is set), so it starts on boot; this PR fixes the unseal half.

## Verified on host (proven end-to-end, not just enabled)
```
[1] systemctl --user is-enabled keiracom-vault-unseal.service  -> enabled
    default.target.wants/keiracom-vault-unseal.service symlink present
[2] docker inspect keiracom-vault  -> RestartPolicy=unless-stopped  Cmd=[server]  running
[3] seal-test (simulates boot):
    sealed=True  -> systemctl --user restart keiracom-vault-unseal.service (re-runs fresh)
    sealed=False -> cold-resolve readback anthropic/api_key len=108  (functional)
```
Note: a manual `systemctl start` on an already-active `RemainAfterExit=yes` oneshot is a no-op; `restart` (or a real boot) re-runs ExecStart. Verified via `restart`.

## Test plan
- [x] `is-enabled` = enabled + boot symlink present
- [x] restart policy = unless-stopped
- [x] controlled seal → unit → unsealed + functional readback
- [ ] Full reboot-test (P11) — this PR is the prerequisite

### DIFF
--- a/infra/systemd/agents/keiracom-vault-unseal.service
+++ b/infra/systemd/agents/keiracom-vault-unseal.service
@@ -1,18 +1,26 @@
 [Unit]
 Description=Auto-unseal the persistent Keiracom Vault on boot (P10 / bd lmce)
 Documentation=file:///home/elliotbot/clawd/Agency_OS/docs/runbooks/vault_persistence.md
-After=docker.service network-online.target
+# NB: docker.service is a SYSTEM unit and invisible to this --user manager, so we
+# do NOT order against it (it would be silently dropped). The boot race against the
+# container is handled instead by the seal-status poll loop inside the ExecStart
+# script + Restart=on-failure below.
+After=network-online.target
 Wants=network-online.target
+StartLimitIntervalSec=300
 
 [Service]
 Type=oneshot
 RemainAfterExit=yes
 Environment=VAULT_ADDR=http://127.0.0.1:8200
-ExecStart=/home/elliotbot/clawd/Agency_OS/scripts/vault_auto_unseal.sh
+# Stable installed copy — deliberately NOT a git-worktree path. /home/elliotbot/clawd/
+# Agency_OS is shared by Elliot/Atlas and is frequently checked out onto feature
+# branches that don't contain this script, which would make a boot unseal fail.
+# install_vault_persistence.sh copies the canonical scripts/vault_auto_unseal.sh here.
+ExecStart=/home/elliotbot/.config/agency-os/vault_auto_unseal.sh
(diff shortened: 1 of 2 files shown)


Keiracom:
[REVIEW:approve:elliot] impl-feasibility: vault boot-resume — keiracom-vault-unseal.service (enabled, verified) starts the container if down AND unseals it on boot, via a stable worktree-independent script path (/home/elliotbot/.config/agency-os/vault_auto_unseal.sh), Restart=on-failure + seal-status poll. Correctly handles container-start + unseal in one unit (no separate restart-policy needed). This closes the vault-reboot gap (P11). Needs a 2nd concur.

Keiracom:
[REVIEW:approve:max] — diff is correct. After=docker.service removal is right (system unit invisible to --user manager; poll loop inside the script is the right way to handle the boot race). StartLimitIntervalSec moved to [Unit] is spec-correct. ExecStart path change to stable ~/.config/agency-os/ copy is sound — the comment in the service file explains the worktree-branch hazard clearly. install_vault_persistence.sh idempotency guards (chmod||true, vault.hcl skip-if-present) are correct for re-run. Gate verified: `systemctl --user is-enabled keiracom-vault-unseal.service` = enabled.

PR #1304, opened 2026-05-29
[ATLAS] fix(work-loop): bridge unit session-DSN override for asyncpg LISTEN (Agency_OS-nkc0)

Keiracom:
## Summary
- The work-loop bridge `LISTEN`s on the Postgres `task_event` channel, but the `.env` DSN is the Supabase **transaction pooler (`:6543`)**, which silently drops `LISTEN` (Supavisor hands a different backend per transaction, so the registration is never pinned → notifies never delivered).
- Adds a **bridge-scoped** `EnvironmentFile` pointing at a session-mode DSN (`:5432`) override file, so `asyncpg LISTEN` works. Isolated to this unit only — the global txn pooler stays the default for every other service.
- The override file (`/home/elliotbot/.config/agency-os/work-loop-bridge.env`) holds the secret **off-repo**; this change is **path-only, no secrets**.
- Makes the live-smoke fix durable: without it, a future `install_work_loop.sh` re-run would revert the installed unit to the broken txn-pooler DSN.

## Context
Found + fixed during the nkc0 work-loop live smoke (2026-05-29). After this fix + applying the kei45 trigger migration to prod, the **full end-to-end smoke PASSED**: `INSERT public.tasks(status=available)` → kei45 trigger → bridge `LISTEN`(session) → Valkey publish → consumer admit (tier-gate/lock/lease) → `/dispatcher/spawn` → env-`i` scrubbed-tmux spawn (P10: bootstrap present, creds scrubbed) → clean agent exit.

## Test plan
- [x] Proved `:5432` session pooler delivers a notify via `asyncpg add_listener` + `pg_notify` round-trip; `:6543` does not.
- [x] Installed unit with this override → bridge log `work-loop bridge LISTENing on PG 'task_event'` + `published a new_available task`.
- [x] Full loop smoke spawned a scrubbed agent that exited clean.
- [ ] Reviewer: confirm later `EnvironmentFile` wins for `SUPABASE_DB_DSN` (systemd ordering) and the `-` prefix keeps an absent file non-fatal.

Note: real-agent go-live is still gated on `agent_cold_start` (Agency_OS-yhm8); the smoke used a vault-resolve probe.

### DIFF
--- a/systemd/keiracom-work-loop-bridge.service
+++ b/systemd/keiracom-work-loop-bridge.service
@@ -19,6 +19,13 @@ StartLimitIntervalSec=300
 [Service]
 Type=simple
 EnvironmentFile=/home/elliotbot/.config/agency-os/.env
+# Bridge-only override: a SESSION-mode DSN (:5432) so asyncpg LISTEN/NOTIFY works.
+# The Supabase TRANSACTION pooler (:6543, the .env default) silently drops LISTEN —
+# Supavisor hands a different backend per txn, so the registration is never pinned.
+# Later EnvironmentFile wins, so SUPABASE_DB_DSN here overrides the .env one for THIS
+# unit only; the global txn pooler stays the default for every other service.
+# Optional (leading '-'): absent file is non-fatal. Holds no secret — only a path.
+EnvironmentFile=-/home/elliotbot/.config/agency-os/work-loop-bridge.env
 WorkingDirectory=/home/elliotbot/clawd/Agency_OS
 ExecStart=/usr/bin/python3 -B -m src.keiracom_system.work_loop.bridge
 


Keiracom:
[REVIEW:approve:elliot] Implementation feasibility: PASS. Minimal and correct — one EnvironmentFile line with leading '-' (non-fatal if absent, right pattern for optional overrides). Later file wins so SUPABASE_DB_DSN overrides only for this unit; global txn pooler untouched. No secret in the file itself. Solves the asyncpg LISTEN pinning issue without blast radius to other services.

Keiracom:
[REVIEW:approve:max]

**Code quality lens — APPROVE**

Both technical questions from the review brief answered and verified against `man systemd.exec`:

**`-` prefix:** Correct. Man page verbatim: *"optionally prefixed with '-', which indicates that if the file does not exist, it will not be read and no error or warning message is logged."* Absent file is non-fatal — unit starts with the base `.env` DSN (txn pooler), LISTEN silently drops, but no crash. Right trade-off: a hard-mandatory file would fail fresh deploys before the ops team creates the override.

**Later EnvironmentFile wins:** Confirmed. Man page verbatim: *"If the same variable is set twice from these files, the files will be read in the order they are specified and the later setting will override the earlier setting."* Ordering in the unit is correct — base `.env` first (`:6543`), override last (`:5432`). `SUPABASE_DB_DSN` resolves to session pooler for this unit only. ✓

**Pattern:** Correct and standard. Per-unit override for a service with a non-default infra requirement. Secret stays off-repo; this PR is path-only. Multiple `EnvironmentFile` directives are explicitly allowed by systemd.

**Comments:** Load-bearing — explain WHY (Supavisor hands a different backend per txn, so LISTEN is never pinned) and the ordering invariant. Not noise.

**Edge case noted:** `-` suppresses ENOENT only; an existing-but-unreadable file would still fail the unit. Standard ops hygiene, not a PR concern — same caveat applies to the first `EnvironmentFile` already in the unit.

**CI:** all pass (Backend Tests 4m53s, Ruff, MyPy, CodeQL). Vercel failures are build-rate-limit, unrelated.

LGTM — merge when Aiden concurs.

PR #1276, opened 2026-05-28
[ATLAS] feat(work-loop): wire release_slot into dispatcher exit hook + periodic reconcile (Agency_OS-innu)

Keiracom:
## Summary

The **#1275 follow-up** (Agency_OS-innu) — closes the self-driving loop end-to-end by wiring the consumer's `release_slot` + `reconcile` into the dispatcher lifecycle.

> **Stacked on #1275.** Base is `atlas/work-loop-consumer` so the diff is isolated to this PR's changes. **Merge order: #1275 first, then this** (GitHub auto-retargets this PR to `main` once #1275 merges).

## What it does

1. **Exit hook** — `/dispatcher/terminate` now frees the tenant slot. The spawn registry retains `tenant_id` (from `spawn_kwargs`); on terminate, after the bounded-spawn release, it calls `work_loop.release_on_exit(tenant_id, key)` → `consumer.release_slot` → **DECR `active_spawns` → LPOP overflow → spawn next**.
2. **Periodic reconcile** — a **Step-6 lifespan background task** runs `consumer.reconcile_all()` every 60s (`SCAN active_spawns:* → reconcile each tenant`) to reclaim slots whose lease TTL lapsed — the **crashed-agent** path the clean exit hook can't cover. Cancelled cleanly on shutdown alongside watchdog/reaper.

## Changes

- **New** `src/keiracom_system/work_loop/integration.py` — `release_on_exit` + `reconcile_loop` + a process-singleton consumer with a `set_consumer` test seam. **Both entry points fail-open** so a work-loop error never breaks dispatcher teardown or the lifespan.
- `consumer.reconcile_all()` — SCAN-based multi-tenant crash sweep (no caller-maintained tenant list; the `active_spawns:{tenant}` keys *are* the active set).
- `dispatcher/main.py` — 3 surgical edits: import, store `tenant_id` in `_spawned`, call `release_on_exit` on terminate + start the Step-6 reconcile task.

## Reuse / scope

Reuses `dispatcher/valkey_pool` + the real `SpawnRequest`/`_spawned` contract (per #1275). I touched the shared `dispatcher/main.py` (exit hook + lifespan) — minimal + flagged. The reaper-driven (TTL) reap path is covered by `reconcile` via lease expiry rather than a per-reap callback (kept scope tight).

## Test plan

[...]

PR #1320, opened 2026-05-29
[MAX] test(dispatcher): regression guard for Agency_OS-28ai reconcile_loop CI hang

Keiracom:
## Summary

- Adds `test_reconcile_loop_not_started_by_default` to `test_main_wiring.py`
- Verifies `work-loop-reconcile` task is NOT created when `DISPATCHER_WORK_LOOP_RECONCILE_ENABLED` is unset (the off-by-default gate)
- Control assertions: watchdog + reaper DO start (unchanged behavior)

## Root cause (Agency_OS-28ai)

`reconcile_loop` was started unconditionally in the dispatcher lifespan (PR #1276). In CI TestClient tests, the loop subscribed to a Valkey channel that never received messages — the loop blocked indefinitely — pytest hung for 27 minutes until the job was force-killed.

The gate (`work_loop_reconcile_enabled=False` by default) was added to `main.py` as the fix with a comment referencing `28ai`. But no test proved the gate held — a future accidental env change could re-enable the hang silently.

This test closes that gap: if the gate breaks, CI fails fast (< 1s) instead of hanging for 27 minutes.

## Test plan
- [ ] `pytest tests/dispatcher/test_main_wiring.py::test_reconcile_loop_not_started_by_default` — 1 passed in 0.50s

### DIFF
--- a/tests/dispatcher/test_main_wiring.py
+++ b/tests/dispatcher/test_main_wiring.py
@@ -287,6 +287,55 @@ def test_terminate_removes_session_from_supervisors(
 # ---------------------------------------------------------------------------
 
 
+# ---------------------------------------------------------------------------
+# Test N — work-loop reconcile gate (Agency_OS-28ai regression guard)
+#
+# Root cause: reconcile_loop was started unconditionally in the lifespan,
+# connecting to an absent Valkey in CI TestClient tests → 27-min hang.
+# Gate: DISPATCHER_WORK_LOOP_RECONCILE_ENABLED=off by default.
+# This test verifies the gate holds: 'work-loop-reconcile' task must NOT
+# be created when the env var is absent.
+# ---------------------------------------------------------------------------
+
+
+def test_reconcile_loop_not_started_by_default(monkeypatch: pytest.MonkeyPatch) -> None:
+    """reconcile_loop must NOT start when DISPATCHER_WORK_LOOP_RECONCILE_ENABLED is unset.
+
+    Regression guard for Agency_OS-28ai: the loop blocked on an absent Valkey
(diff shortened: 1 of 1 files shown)


Keiracom:

Gate confirmed at main.py:613 (DISPATCHER_WORK_LOOP_RECONCILE_ENABLED, off by default). Test is a regression guard — not testing new behavior, locking existing behavior so it can't regress silently.

- `importlib.reload(main_mod)` is the correct approach here: the gate is evaluated at module import time (line 255-256), not inside a function — so testing it requires a fresh import with the env var unset ✓
- Task-name spy via asyncio.create_task is clean and won't produce false positives ✓  
- Control assertions (watchdog + reaper DO start) verify non-regression on other tasks ✓
- Root cause documented in comment ✓

Max authored → Aiden eligible for 2-of-3. Pinging Aiden.

Keiracom:
[REVIEW:approve:aiden]

**Pytest failure:** `test_tasks_cli_evidence.py` — confirmed pre-existing on `main` (run 26641659294 fails the same tests on the latest main push). Not introduced by this PR.

**Gate pattern:** `DISPATCHER_WORK_LOOP_RECONCILE_ENABLED` off-by-default is consistent with the dispatcher's established pattern — `_idempotency_gate`, `_budget_gate`, `_cost_breaker` are all `None`/disabled by default and wired at runtime. A Valkey-dependent loop following the same discipline is architecturally correct.

**GOV-12:** The test IS the runtime enforcement — `asyncio.create_task` spy + `importlib.reload` verifies the gate holds at lifespan time. Control assertions (watchdog + reaper start; reconcile does not) make it non-trivially falsifiable. If the gate is removed from `main.py`, this test catches it.

PR #1319, opened 2026-05-29
[ATLAS] feat(ops): OnFailure= alerting → NATS keiracom.ops.failure → #ceo (Agency_OS-ja8d)

Keiracom:
## Summary
Closes the silent-stale blind spot (Agency_OS-ja8d): failed user-scope services currently emit **zero** alert. The KEI-141 poller is **not running** (timer/service inactive, not-installed) and even when running only covered 5 unit-globs. `weaviate-backup.service` was failing every boot unnoticed — the live instance of this gap.

This wires **push-based** `OnFailure=` alerting:

```
failed unit ──OnFailure──▶ keiracom-ops-failure-alert@<unit>.service
                              └─▶ ops_failure_publish.py → NATS keiracom.ops.failure
                                    └─▶ peer_event_ceo_relay (extended) → Slack #ceo
```

## Files
- **`scripts/orchestrator/ops_failure_publish.py`** (new) — gathers `systemctl status` + `journalctl` for the failed unit, publishes a diagnostic envelope to `keiracom.ops.failure`. Core NATS pub/sub, mirrors `supervisor_wake_publish.py` (lazy `import nats`, fail-open diagnostics).
- **`infra/systemd/agents/keiracom-ops-failure-alert@.service`** (new) — oneshot handler template. **Uses `%i`, not `%I`** — verified `%I` unescapes `-`→`/` (`weaviate-backup.service` → `weaviate/backup.service`), which would corrupt every dashed unit name.
- **`scripts/install_onfailure_dropins.sh`** (new) — idempotent installer; drops `OnFailure=keiracom-ops-failure-alert@%n.service` on every concrete `*.service`; `--remove` reverses it; excludes the alert template (recursion guard).
- **`scripts/orchestrator/peer_event_ceo_relay.py`** (edit) — **minimal** diff: +`keiracom.ops.failure` subscription, +`_handle_ops_failure` (per-unit throttle) → `_post_ceo("Service failure", …)`.
- **`tests/orchestrator/test_ops_failure_alerting.py`** (new) — 8 tests.

## Scope notes for reviewers
- **GOV-9 correction:** the original ja8d spec said *"peer-event-ceo-relay then fans → #ceo"* — but the relay did **not** subscribe to `keiracom.ops.failure` (only `elliot.inbox` + `review.>`). That missing subscription is part (c) here.
[...]

PR #1325, opened 2026-05-29
[ATLAS] fix(ops): cover @ template instances in OnFailure drop-ins (Agency_OS-cws5)

Keiracom:
## Summary
ja8d follow-up (closes the gap filed in Agency_OS-cws5): `install_onfailure_dropins.sh` skipped `*@.service` templates, so the 7 running `agent-self-claim-loop@<callsign>` instances had **no** OnFailure alert.

## Why templates are safe to cover
A drop-in in `foo@.service.d/` applies to every instance `foo@<x>.service`. The failing instance's full unit name survives the chain:
`%n` (full unit name) → handler `keiracom-ops-failure-alert@%n.service` → `%i` (verbatim — **not** `%I`, which unescapes `-`→`/`).

**Verified live** (`--selftest`): `zz-onfail-selftest@probe.service` failed → publisher logged `unit=zz-onfail-selftest@probe.service` (double-`@` + dashes intact, no corruption).

## Changes
- Remove the `*@.service` skip — templates now covered via their `.d` dir (the script already computed `<unit>.d` correctly).
- Keep the `keiracom-ops-failure-alert@` exclusion (recursion guard).
- Add `--selftest`: a self-contained live double-`@` regression test (creates a throwaway template, fails an instance, asserts the publisher saw the correct instance unit name, cleans up). Exits 0/1.

## Verification (live, deployed host-side — reversible)
```
$ bash scripts/install_onfailure_dropins.sh --selftest
SELFTEST PASS: alert fired with correct instance unit 'zz-onfail-selftest@probe.service'

$ bash scripts/install_onfailure_dropins.sh
done — wrote 100 drop-in(s), skipped 1 unit(s)   # was: 99 wrote / 2 skipped

$ systemctl --user show agent-self-claim-loop@atlas.service -p OnFailure
OnFailure=keiracom-ops-failure-alert@agent-self-claim-loop@atlas.service.service
```
`bash -n` clean. Alert template still excluded (no self-drop-in → no recursion).

## Test plan
- [ ] `--selftest` passes on a host with the alert template + NATS deployed
- [ ] Fail a real agent-self-claim-loop instance → alert in #ceo with correct unit name
- [ ] `--remove` cleanly removes template drop-ins too

### DIFF
--- a/scripts/install_onfailure_dropins.sh
+++ b/scripts/install_onfailure_dropins.sh
@@ -8,16 +8,22 @@
 #   [Unit]
 #   OnFailure=keiracom-ops-failure-alert@%n.service
 #
+# Template units (foo@.service) ARE covered (Agency_OS-cws5): the drop-in lands
+# in foo@.service.d/, which systemd applies to every instance foo@<x>.service.
+# When an instance fails, %n is its full unit name; the handler
+# keiracom-ops-failure-alert@%n.service passes it verbatim to the publisher via
+# %i (NOT %I — %I unescapes '-' to '/'). The resulting double-@ instance name
+# resolves correctly — verified live via --selftest.
+#
 # EXCLUSIONS (never add a drop-in to):
-#   - keiracom-ops-failure-alert@.service (the template itself)
-#   - keiracom-ops-failure-alert@*.service instances (prevents recursion)
-#   - Any template unit whose name contains '@' with empty instance (e.g. foo@.service)
+#   - keiracom-ops-failure-alert@.service + its instances (prevents recursion)
 #
 # Idempotent: re-running writes the same content without error.
 #
 # Usage:
-#   bash scripts/install_onfailure_dropins.sh           # install
-#   bash scripts/install_onfailure_dropins.sh --remove  # remove all drop-ins
+#   bash scripts/install_onfailure_dropins.sh            # install
+#   bash scripts/install_onfailure_dropins.sh --remove   # remove all drop-ins
+#   bash scripts/install_onfailure_dropins.sh --selftest # live double-@ test
 set -euo pipefail
 
 REPO_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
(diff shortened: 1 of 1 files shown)


Keiracom:

- Template coverage: drop-in in foo@.service.d/ correctly applies to all instances ✓
- %n → keiracom-ops-failure-alert@%n.service → %i chain documented — instance full unit name survives with @ intact ✓
- %i not %I reaffirmed in comment (regression guard for the earlier %I bug) ✓
- --selftest flag for repeatable live regression test — the right way to verify double-@ behavior without relying on a real service failure ✓
- 100 drop-ins (was 99), only recursion guard skipped ✓
- Deployed + live-verified by Atlas (PASS with zz-onfail-selftest@probe.service) ✓

Needs one deliberator concur — pinging Max (Aiden is on zr7e.2 design).

Keiracom:
[REVIEW:approve:max] Code quality lens — APPROVE.

**Recursion guard simplification:** Two exclusion conditions collapsed to one glob: `keiracom-ops-failure-alert*`. Covers the template AND all running instances in a single match. Cleaner and less likely to miss an edge case than separate pattern checks.

**Template drop-in semantics:** Removing the `*@.service` skip is the correct fix. A drop-in in `foo@.service.d/` applies to every instance `foo@<x>.service` — that's the designed systemd mechanism for template-wide policy. The original exclusion was over-broad and silently left all agent self-claim-loop instances unmonitored.

**Double-@ chain:** `%%n` in the printf statement outputs literal `%n` to the conf file — same as the heredoc `<<'DROPIN'` in the main install path. systemd expands `%n` to the failing instance's full unit name (`zz-onfail-selftest@probe.service`). `%i` (not `%I`) in the handler template passes it verbatim — no '-'→'/' corruption. Consistent with the PR #1319 decision already anchored in comments.

**--selftest:** Start/fail/sleep-4/grep/cleanup sequence is correct for a live smoke test. `reset-failed` before removal is important (stops systemd retaining the unit in failed state). `|| true` guards on the start call and the rmdir are correct (start returns non-zero on deliberate failure; rmdir only needed if dir is now empty).

Atlas authored → Elliot + Max = 2-of-3. Merge eligible.

PR #1312, opened 2026-05-29
[MAX] feat(work-loop): result-back-to-Slack — task_complete endpoint + cold_start notify

Keiracom:
## Summary

- Adds `POST /dispatcher/task_complete` endpoint to `src/dispatcher/main.py` — receives `{task_id, callsign, title, status, rc}` from worker agent on completion
- `notify_complete()` injected into `agent_cold_start.py` run() after `finalize_task()` — urllib-only stdlib POST, fail-open
- Slack relay bridged via subprocess with `CALLSIGN=elliot` (host env holds `SLACK_BOT_TOKEN`; scrubbed agent env does not)
- 14 new tests (9 cold_start + 5 dispatcher), 24 total passing, 0 broken existing

## Test plan

- [ ] `pytest tests/dispatcher/test_task_complete.py` — 5 endpoint tests pass
- [ ] `pytest tests/keiracom_system/vault/test_agent_cold_start.py` — 9 new + existing pass
- [ ] Spin up dispatcher + spawn a cold_start agent → verify Slack message lands in #ceo

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1329, opened 2026-05-29
[ATLAS] feat(chain): v1_chain_orchestrator — first-hop dispatch + state machine (Agency_OS-zr7e.2)

Keiracom:
## Summary
V1 core: per-task chain-state tracker that routes a Face-classified task through `face → aiden → max → nova(worker) → orion+atlas(reviewers) → Slack` via NATS dispatch + an `advance_step` API. **First-hop ACCEPTANCE verified live.**

## Architecture (from my own zr7e.2 research, ratified by Elliot)
- **Dispatch transport = NATS** (per-callsign bridges already enabled fleet-wide).
- **Workers** subscribe `keiracom.dispatch.<self>`; **deliberators** subscribe `keiracom.deliberation.>`.
- **Worker-first for V1 = Nova** (Orion=Aiden's worker but acts as post-Worker spec-reviewer in the V1 chain — using Orion as Worker would be self-review).
- **Atom_id passing:** canonical envelope key `atom_id` (coordination point with Orion's `zr7e.9` — they provide the NATS publish side, this orchestrator subscribes).

## Files
- `src/keiracom_system/chain/__init__.py` — new (package)
- `src/keiracom_system/chain/v1_chain_orchestrator.py` — new (270 LoC) — `dispatch()` + `advance_step()` + state-file persistence + NATS publish helper (mirrors `supervisor_wake_publish`, fail-open).
- `src/keiracom_system/chat/face.py` — `_respond` wired to call `chain_dispatch` on non-ambiguous classification (fail-open: any exception → stub reply, Face never propagates). Added `brief: str = ""` kwarg (backward-compatible).
- `tests/keiracom_system/chain/test_v1_chain_orchestrator.py` — 8 tests (first-hop, fail-open, state persistence, sequential aiden→max→nova, parallel nova→orion+atlas, pending-list gating, dual-reviewer completion).
- `tests/keiracom_system/chat/test_face.py` — 3 new face tests (dispatches on classified, ambiguous unchanged, fail-open if dispatch raises).

## Envelope (canonical)
```json
{
  "task_id": "...", "chain_id": "...", "chain_step": "aiden_plan|max_challenge|nova_build|orion_spec|atlas_safety",
  "atom_id": null | "...", "brief": "...", "ts": <float>, "from": "v1_chain_orchestrator"
}
```
Publish subject: `keiracom.dispatch.<role>` (role from `ROLE_FOR_STEP`). State JSON at `/tmp/v1_chain_state.json` (env-overridable, atomic write).

[...]

PR #1337, opened 2026-05-29
[NOVA] fix(notify_complete): suppress intermediate chain-step #ceo posts (Agency_OS-nd3b)

Keiracom:
## Summary
Closes Agency_OS-nd3b. V1 chain hops (Face → Aiden → Max → Worker → Orion + Atlas) each call `task_complete`, which posts to **#ceo** via PR #1312's dispatcher → `slack_relay` path. Pre-fix Dave would see **N notifications per directive**; post-fix he sees **ONE** — the final reviewer.

## Change
Gate at the top of `notify_complete` reads the `CHAIN_STEP` env var (injected by the dispatcher from the `v1_chain_orchestrator` envelope at spawn):

| `CHAIN_STEP` value | behavior |
|---|---|
| absent | POST as today (backward-compatible — legacy single-step tasks notify) |
| `'aiden_plan'` / `'max_challenge'` / `'nova_build'` / `'orion_review'` / `'atlas_review'` | log `suppressed` + **return** (no POST) |
| `'complete'` | POST (final step) |

Fail-open path is unchanged — the suppression is a clean early-return before any network work.

## Test plan
- [x] 3 new tests: intermediate suppressed (urlopen never called + 'suppressed' log line); `'complete'` posts; CHAIN_STEP absent posts.
- [x] `pytest tests/keiracom_system/vault/test_agent_cold_start.py` → **34 passed** (31 existing + 3 new).
- [x] `ruff check` clean.

### DIFF
--- a/src/keiracom_system/vault/agent_cold_start.py
+++ b/src/keiracom_system/vault/agent_cold_start.py
@@ -189,7 +189,22 @@ def notify_complete(
 
     Fail-open: any error (network, dispatcher down, Slack failure) is logged
     and swallowed — a notification failure must never block the task lifecycle.
+
+    Chain-step suppression (Agency_OS-nd3b): the v1_chain_orchestrator dispatch
+    envelope carries `chain_step` and the dispatcher injects it as the CHAIN_STEP
+    env var at spawn. Only the FINAL step (CHAIN_STEP=='complete') posts to #ceo
+    — intermediate hops (aiden_plan / max_challenge / nova_build / orion_review /
+    atlas_review) suppress to keep Dave from seeing N notifications per directive.
+    CHAIN_STEP absent preserves today's behavior (single-step legacy tasks notify).
     """
+    chain_step = os.environ.get("CHAIN_STEP")
+    if chain_step and chain_step != "complete":
+        logger.info(
+            "notify_complete: suppressed (intermediate chain_step=%s) task=%s",
+            chain_step,
+            task_id,
+        )
+        return
     payload = json.dumps(
         {"task_id": task_id, "callsign": callsign, "title": title, "status": status, "rc": rc}
     ).encode()
--- a/tests/keiracom_system/vault/test_agent_cold_start.py
+++ b/tests/keiracom_system/vault/test_agent_cold_start.py
@@ -239,6 +239,82 @@ def test_notify_complete_fail_open_on_network_error(monkeypatch):
     # no exception = pass
 
 
+# ---- chain-step suppression (Agency_OS-nd3b) -------------------------------
+
+
+def test_notify_complete_suppressed_when_chain_step_is_intermediate(monkeypatch, caplog):
+    """CHAIN_STEP set to an intermediate step → no POST (urlopen never called)."""
+    import logging
+    import urllib.request as urlreq
+
+    called: list = []
+
+    def boom_urlopen(_req, timeout):
+        called.append(True)
+        raise AssertionError("urlopen MUST NOT be called for intermediate chain steps")
+
+    monkeypatch.setenv("CHAIN_STEP", "aiden_plan")
+    monkeypatch.setattr(urlreq, "urlopen", boom_urlopen)
+    with caplog.at_level(logging.INFO, logger="src.keiracom_system.vault.agent_cold_start"):
+        acs.notify_complete("t-int", "aiden", "Plan KEI-1", "done", 0)
+    assert called == []  # urlopen never invoked
+    assert any("suppressed" in rec.message and "aiden_plan" in rec.message for rec in caplog.records)
+
+
+def test_notify_complete_posts_when_chain_step_is_complete(monkeypatch):
+    """CHAIN_STEP='complete' (final step) → posts as normal."""
+    import urllib.request as urlreq
+
+    captured: dict = {}
+
+    class _FakeResp:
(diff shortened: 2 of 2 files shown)


Keiracom:

Gate logic is exactly right:
- `chain_step = os.environ.get('CHAIN_STEP')` — reads env at call time ✓
- `if chain_step and chain_step != 'complete': return` (suppress) — correct: absent=legacy-compat, intermediate=suppress, complete=post ✓
- Three-way test coverage: intermediate suppressed (urlopen never called + suppressed log), complete posts, absent posts ✓
- 34 vault tests green — no regression in existing notify paths ✓
- Backward-compatible: legacy single-step tasks (no CHAIN_STEP) notify as before ✓
- Dispatcher must inject CHAIN_STEP=<step> at spawn — noted in docstring ✓

Dave now sees one #ceo notification per chain run (final reviewer step) instead of N.

Keiracom:
[REVIEW:approve:max] Code quality + test coverage lens — APPROVE.

**3-case gate: correct and sufficient**

```python
chain_step = os.environ.get('CHAIN_STEP')
if chain_step and chain_step != 'complete':
    return  # suppress
```

Three logical branches:
1. Absent / empty-string: `if chain_step` is False → falls through → posts. Backward compat preserved. ✓
2. `'complete'`: `chain_step != 'complete'` is False → falls through → posts. ✓
3. Any other non-empty value (all intermediate steps): `if chain_step` is True and not 'complete' → suppressed. ✓

Any unrecognised CHAIN_STEP value (e.g. typo 'complet') suppresses rather than posts. Fail-safe direction — better to miss a notification than to spam Dave with an unknown step. ✓

**Edge case: 'complete' mis-injected on an intermediate step**

The gate cannot defend against a wrong value from the dispatcher. If the orchestrator's dispatch envelope writes chain_step='complete' for an intermediate step, the gate posts. But this is a dispatcher-injection contract concern, not a gate logic flaw. The v1_chain_orchestrator correctly sets chain_step per step (verified in PR #1329 review) — 'complete' only appears in the terminal state. The gate is correct given the contract.

**`urlopen-never-called` assertion: correct pattern**

`boom_urlopen` appends to `called` BEFORE raising AssertionError. Since notify_complete's outer `except Exception` swallows the AssertionError, `called == []` is the reliable detection mechanism — the raise message is informative in un-swallowed contexts, the list check is what actually catches suppression failure. ✓

The caplog assertion (`'suppressed' in message and 'aiden_plan' in message`) verifies the suppression *path* was taken, not just that urlopen was skipped due to some other early return. Two independent confirmation signals. ✓

**Tests: 3-for-3**

- Intermediate (aiden_plan): urlopen never called + log verified. ✓
- Final (complete): urlopen called, payload verified. ✓
- Absent: urlopen called, URL verified (backward compat). ✓

All 3 branches exercised, no coverage gap.

Nova authored → Elliot + Max = 2-of-3. Merge eligible.

Keiracom:
[REVIEW:approve:aiden]

**CHAIN_STEP env var correctly models chain state:** the chain envelope's `chain_step` field is the canonical position signal, and the dispatcher injecting it as env at spawn is the correct decoupling. The agent has no other way to know its chain position without querying Postgres — that would couple agent_cold_start to the chain orchestrator table and violate the task-layer-as-black-box constraint from `ceo:v1_chain_orchestrator_design`. ✓

**Dispatcher-at-spawn is the right coupling point:** the dispatcher is the only party that simultaneously holds the chain envelope and the spawn context. Env injection is the correct direction. ✓

**Non-chain task suppression risk is low:** gate condition `if chain_step and chain_step != 'complete'` evaluates `if None → False → POST` for any task without CHAIN_STEP. Legacy/non-chain tasks are unaffected by construction. P2 NIT: CHAIN_STEP is not namespaced (`KEIRACOM_CHAIN_STEP` would be more defensive against env contamination), but this is not a blocker. ✓

**Intermediate failure suppression is correct for V1:** a chain agent with rc != 0 is also suppressed (intermediate failure). Dave doesn't see N blocked notifications per directive — the chain orchestrator owns failure state via `chain_advance` and the `complete` step carries the final verdict. ✓

**Tests:** all three branches covered (intermediate → suppress + log, complete → POST, absent → POST legacy compat). ✓

Keiracom:
[REVIEW:HOLD:aiden] — REVISING prior approval. Missed a design error.

**HOLD: `CHAIN_STEP='complete'` is never injected — gate permanently suppresses all chain task notifications.**

`complete` is a terminal state in `v1_chain_orchestrator`, stored in `entry["current_step"]` when the chain finishes. But no agent is ever SPAWNED with `CHAIN_STEP=complete`: `ROLE_FOR_STEP` has no `complete` entry, and `advance_step()` transitions to `complete` state without dispatching another agent. So the condition `chain_step != 'complete'` is always `True` for every valid dispatched value (`aiden_plan`, `max_challenge`, `nova_build`, `orion_spec`, `atlas_safety`) — the gate permanently suppresses every chain task notification, including `atlas_safety` (the final step). Dave never sees any chain task complete in #ceo.

**Resolution:** adopt Orion's approach — `_FINAL_CHAIN_STEPS = frozenset({"atlas_safety"})` names the last ACTUAL dispatched step. PR #1338 implements this correctly. Recommend closing #1337 in favour of #1338.

PR #1338, opened 2026-05-29
[ORION] feat(cold-start): suppress notify on intermediate V1-chain hops (Agency_OS-nd3b)

Keiracom:
## Summary
`agent_cold_start` posts every task completion to `#ceo` via `notify_complete`. For V1-chain hops (Aiden plan → Max challenge → Nova build → Orion spec + Atlas safety), that would spam `#ceo` with per-hop completions; only the final reviewer should announce. **nd3b** adds a `CHAIN_STEP` env-var gate.

- **`_FINAL_CHAIN_STEPS = {"atlas_safety"}`** — mirrors the canonical taxonomy from `src.keiracom_system.chain.v1_chain_orchestrator.CHAIN_STEPS` (`aiden_plan, max_challenge, nova_build, orion_spec, atlas_safety`). Per `ceo:v1_chain_architecture`, Atlas reviews **after** Orion confirms spec — the Atlas persona's first instruction reads *"Assume Orion has confirmed the output matches the KEI"* — so `atlas_safety` is the **last step in time** and the only one whose completion should post `#ceo` from this entrypoint.
- **`run()` wraps `notify`** in a `CHAIN_STEP` gate: if env is set **and** value is **not** in `_FINAL_CHAIN_STEPS` → log + skip; otherwise notify as today. `CHAIN_STEP` unset / whitespace-only → notify (fail-open default — non-chain tasks + manual invocations are unaffected).
- The final dual-concur result post is **zqni's** territory (which `nd3b` blocks).

## Foundation
- ✅ `git log --all --oneline | grep -iE 'nd3b|notify.*suppress|chain.*step.*flag|skip.*notify|intermediate.*chain'` — nothing shipped.
- ✅ `v1_chain_orchestrator.CHAIN_STEPS` already on `main` (#1329) — labels mirrored here.

## Test plan
- [x] `ruff check` + `ruff format --check` clean
- [x] `pytest tests/keiracom_system/vault/test_agent_cold_start.py` → **35 passed** (31 existing preserved + 4 new for nd3b)
- [x] All 4 intermediate steps (`aiden_plan`, `max_challenge`, `nova_build`, `orion_spec`) suppress notify
- [x] `CHAIN_STEP=atlas_safety` fires notify (final-reviewer path)
- [x] `CHAIN_STEP` unset → notify fires (regression guard for non-chain tasks)
- [x] `CHAIN_STEP="   "` → treated as unset → notify fires (strip-based detection)

## Coordination
- Atlas is on **oevr** (consumer loop) per Elliot's routing. nd3b touches only the post-`save_atoms` notify call site in `run()` — no shared files with the consumer-loop work.

### DIFF
--- a/src/keiracom_system/vault/agent_cold_start.py
+++ b/src/keiracom_system/vault/agent_cold_start.py
@@ -67,6 +67,18 @@ RC_TASK_ABSENT = 3
 NATS_URL = os.environ.get("NATS_URL", "nats://127.0.0.1:4222")
 HANDOFF_SUBJECT = "keiracom.agent.handoff"
 
+# nd3b — chain-step taxonomy mirrors src.keiracom_system.chain.v1_chain_orchestrator
+# CHAIN_STEPS (aiden_plan, max_challenge, nova_build, orion_spec, atlas_safety).
+# The chain branches after nova_build → BOTH reviewers in parallel; per the
+# canonical (ceo:v1_chain_architecture) Atlas reviews AFTER Orion confirms spec,
+# so atlas_safety is the last step in time — the only one whose completion
+# warrants the #ceo task_complete post from this entrypoint. All other
+# CHAIN_STEP values are intermediate hops; suppress notify so the chain does not
+# spam #ceo with per-hop completions. (The final dual-concur result post lands
+# via zqni once both reviewers concur.) CHAIN_STEP unset = non-chain task →
+# notify as today (fail-open default; manual / pre-chain dispatches unaffected).
+_FINAL_CHAIN_STEPS: frozenset[str] = frozenset({"atlas_safety"})
+
 
 def _dsn() -> str:
     dsn = os.environ.get("DATABASE_URL") or os.environ.get("SUPABASE_DB_DSN")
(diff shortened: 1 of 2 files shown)


Keiracom:
Superseded by #1337 (Nova) which shipped the same nd3b fix and has full 3-of-3 deliberator concur. No action needed on this PR.

Keiracom:
[REVIEW:approve:aiden]

**`_FINAL_CHAIN_STEPS = frozenset({'atlas_safety'})` is architecturally correct:** `atlas_safety` is the last ACTUAL dispatched step in the V1 chain. `complete` is a terminal state stored in chain_runs — it's never dispatched as an agent. The frozenset approach is also extensible: if the chain topology changes (new final step), add it to the set without changing the gate logic. ✓

**Gate logic is correct:** `if chain_step and chain_step not in _FINAL_CHAIN_STEPS` — non-chain tasks (CHAIN_STEP absent) evaluate `if ''` → False → notify (fail-open default). Whitespace-only CHAIN_STEP stripped to `''` → same result. Intermediate steps suppress. `atlas_safety` notifies. All four branches tested. ✓

**Suppression placement:** gate is in `run()` rather than inside `notify_complete()` — this is a structural difference from PR #1337. Placing the gate in run() means the suppression is visible in the agent lifecycle rather than hidden inside the notify helper. Clean. ✓

**Conflict with PR #1337:** both implement nd3b. PR #1337 has a design error (`complete` is never dispatched; gate permanently suppresses). Recommend Elliot closes #1337 and merges this one.

PR #1341, opened 2026-05-29
[ORION] feat(dispatcher): inject CHAIN_STEP env at spawn from chain envelope (Agency_OS-qjl7)

Keiracom:
## Summary
End-to-end wiring for the V1-chain notify-suppression gate (**qjl7**, Elliot dispatch). `v1_chain_orchestrator` emits envelopes with a `chain_step` field (e.g. `"aiden_plan"`); the consumer hands that through as a top-level `spawn_kwargs` key; until now the dispatcher's translator would land it as **`AGENT_CHAIN_STEP`** (via the generic metadata loop's `AGENT_<KEY>` mapping), which `agent_cold_start` can't read — its nd3b gate reads `CHAIN_STEP` un-prefixed. Result: chain spawns would always notify on every hop, defeating #1337.

- **`_apply_chain_step_env(sk, env)`** — shared helper: pops `sk['chain_step']` and `env.setdefault('CHAIN_STEP', str(value))`. Pop is intentional — if it stayed in `sk`, the metadata loop would also surface it as `AGENT_CHAIN_STEP`. `setdefault` preserves a caller-supplied `env` override.
- Called from **`_container_spawn_kwargs` AND `_tmux_spawn_kwargs`** right after env init and **before** the `AGENT_<KEY>` metadata loop, so both backends end up with the un-prefixed var (container: `env` dict; tmux: `env -i` assignment in the scrubbed command).
- **Fail-open**: `chain_step` absent → no-op → legacy spawns unchanged. The whole change is a no-op for non-chain dispatches.

## Foundations / pre-flight
- ✅ `git log --all --oneline | grep -iE 'chain_step.*env|CHAIN_STEP.*inject|spawn.*chain_step|envelope.*chain_step'` — nothing shipped.
- ✅ `v1_chain_orchestrator._build_envelope` carries `chain_step` (verified on `main`).
- ✅ `agent_cold_start` reads `CHAIN_STEP` un-prefixed (nd3b / #1337, on `main`).
- ✅ Both dispatcher translators rely on the same `for meta_key, meta_val in sk.items(): env.setdefault(f"AGENT_{meta_key.upper()}"…)` pattern — the helper sits just before that loop in both.

## Acceptance
1. ✅ Container spawn: `chain_step` in `spawn_kwargs` → `env['CHAIN_STEP']` (un-prefixed), NOT `AGENT_CHAIN_STEP` (`test_container_spawn_kwargs_promotes_chain_step_to_env`).
2. ✅ Tmux spawn: `env -i` command carries `CHAIN_STEP=<value>` (`test_tmux_spawn_kwargs_promotes_chain_step_into_scrubbed_command`).
3. ✅ Both backends: `chain_step` absent → no `CHAIN_STEP` set (legacy spawn behaviour preserved).

## Test plan
- [x] `ruff check` + `ruff format --check` clean
- [x] `pytest tests/dispatcher/test_main_chain_step_wiring.py` → **8/8 passed** (helper alone × 4 + container path × 2 + tmux path × 2)
- [x] Dispatcher regression: 20/20 across `test_main_container_defaults` + `test_main_scrubbed_tmux` + `test_main_wiring` (one pre-existing local-shell-env failure in `test_scrub_gated_off_by_default` — `DISPATCHER_TMUX_SCRUB_ENABLED=true` is set in my local shell only; CI is clean and the test isn't touched by this change)

## V1 chain coordination
- nd3b (#1337) on main: `agent_cold_start.run()` reads `CHAIN_STEP` and suppresses `notify` on intermediate hops.
- This PR: dispatcher promotes envelope `chain_step` → env `CHAIN_STEP` so nd3b actually sees a value at spawn.
- Atlas's oevr (consumer loop) will translate handoff messages → `/dispatcher/spawn` calls with `chain_step` in spawn_kwargs.

Together these close the loop: chain orchestrator emits → consumer dispatches → dispatcher injects env → `agent_cold_start` reads env → notify-suppression fires correctly.

### DIFF
--- a/src/dispatcher/main.py
+++ b/src/dispatcher/main.py
@@ -421,6 +421,22 @@ def _free_port() -> int:
         return int(sock.getsockname()[1])
 
 
+def _apply_chain_step_env(sk: dict[str, Any], env: dict[str, Any]) -> None:
+    """qjl7 — promote spawn_kwargs['chain_step'] to env['CHAIN_STEP'] (un-prefixed).
+
+    The v1_chain_orchestrator emits envelopes with a ``chain_step`` field (e.g.
+    "aiden_plan"); the consumer hands that through as a top-level spawn_kwargs
+    key. ``agent_cold_start.run()`` reads ``CHAIN_STEP`` (not the AGENT_* form)
+    for its nd3b notify-suppression gate, so the generic metadata loop's
+    AGENT_<KEY> mapping would land it under the wrong name. Pop the key so it
+    is not double-set; ``setdefault`` keeps any caller-supplied CHAIN_STEP
+    override. Fail-open: chain_step absent → no-op; legacy spawns unchanged.
+    """
+    chain_step = sk.pop("chain_step", None)
+    if chain_step is not None:
+        env.setdefault("CHAIN_STEP", str(chain_step))
+
+
 def _container_spawn_kwargs(key: str, sk: dict[str, Any]) -> dict[str, Any]:
     """Translate logical spawn_kwargs → container_lifecycle.spawn_container kwargs.
 
@@ -437,6 +453,7 @@ def _container_spawn_kwargs(key: str, sk: dict[str, Any]) -> dict[str, Any]:
     port = int(sk.pop("port", None) or _free_port())
     health_path = sk.pop("health_path", None)
     extra_args = sk.pop("extra_args", None)
+    _apply_chain_step_env(sk, env)
     for meta_key, meta_val in sk.items():
         if meta_val is not None:
             env.setdefault(f"AGENT_{meta_key.upper()}", str(meta_val))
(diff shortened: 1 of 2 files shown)


Keiracom:
Superseded by Atlas's #1339 + Nova's nd3b fallback (resolved-by-#1339-fix). Per Elliot redirect (2026-05-29 21:48Z): Scout found the dispatcher's existing AGENT_* injection loop (main.py:441-443, 513-515) already maps chain_step → AGENT_CHAIN_STEP. Atlas's #1339 consumer adds spawn_kwargs={'chain_step': step}; a 1-line nd3b fallback reads AGENT_CHAIN_STEP. No dispatcher diff needed. Closing as redundant — both approaches close the loop; the team chose the simpler one. Note: this PR was correct (functionally equivalent) but unnecessary; the dispatch's 'FILE: src/dispatcher/main.py' pointer led me there before Scout's finding landed.

Keiracom:
[REVIEW:approve:aiden]

Architecture/governance lens — APPROVE. This is the missing wiring without which nd3b never fires.

**`_apply_chain_step_env` is the correct fix:** without this, `chain_step` from the envelope flows through the generic AGENT_<KEY> loop and lands as `AGENT_CHAIN_STEP` — not `CHAIN_STEP`. The nd3b gate in `agent_cold_start` reads `CHAIN_STEP` (un-prefixed), so the suppression never fires and every chain hop notifies. Pop + setdefault is the right pattern: pop prevents the double-set, setdefault preserves any caller-supplied override. ✓

**Shared helper called in both backends:** `_container_spawn_kwargs` and `_tmux_spawn_kwargs` both call `_apply_chain_step_env` BEFORE the generic loop. Consistent. ✓

**Fail-open:** chain_step absent in spawn_kwargs → `sk.pop('chain_step', None)` returns None → no-op → legacy spawns unchanged. ✓

**V1 end-to-end closure:** orchestrator emits `chain_step` in envelope → consumer (PR #1339) passes to spawn_kwargs → THIS PR injects `CHAIN_STEP` → agent_cold_start reads it → nd3b suppression fires correctly. ✓

PR #1357, opened 2026-05-30
[ATLAS] feat(v1-chain): __main__ --consumer branch for systemd entry point (Agency_OS-zr7e)

Keiracom:
## Summary
Wires `v1_chain_orchestrator.run_consumer()` to a systemd-runnable entry point so `keiracom.agent.handoff` NATS envelopes are actually consumed and the V1 chain advances. Per Elliot dispatch 2026-05-30, Agency_OS-zr7e / battery unblock.

## Change
`src/keiracom_system/chain/v1_chain_orchestrator.py` — extends the `__main__` block to branch:
- `--consumer` → `asyncio.run(run_consumer())` (long-running NATS subscriber)
- any other argv → existing dispatch smoke (unchanged behavior for ad-hoc test runs)

5 lines net.

## Companion (host-only — not in this PR)
`~/.config/systemd/user/keiracom-v1-chain-consumer.service` per Elliot's verbatim template, installed on the host. Per `orchestrator-merge-after-NATS-concur`: systemd units in `~/.config/` are host-side infra, not PR-scoped. No `OnFailure=` / Drop-In `onfailure.conf` per dispatch.

## Verification (manual)
Ran `python3 -B -m src.keiracom_system.chain.v1_chain_orchestrator --consumer` from this branch's worktree against the live NATS server. Captured journal-equivalent log:

```
2026-05-30 06:45:06,492 INFO v1_chain consumer: subscribed
  keiracom.agent.handoff on nats://127.0.0.1:4222
```

Subscription fires; consumer stays in the `await asyncio.sleep(60)` loop until SIGTERM. Code path is correct.

The systemd unit was initially `enable --now` while still running against `/home/elliotbot/clawd/Agency_OS` (the main worktree, pre-merge stale code). It hit the smoke `dispatch()` else-branch and auto-restarted in a loop. Stopped + disabled until this PR merges; activation chain is then `merge → systemctl --user enable --now keiracom-v1-chain-consumer.service`.

## Test plan
- [x] `ruff format --check` + `ruff check` clean
- [x] `pytest tests/keiracom_system/chain/test_v1_chain_orchestrator.py` — 27 passed (no regressions; `__main__` block is `# pragma: no cover`)
- [x] Manual `--consumer` run from atlas worktree subscribes to `keiracom.agent.handoff`
- [ ] Aiden NATS concur per orchestrator-merge-after-NATS-concur (2-of-2 author-exclusion)
- [ ] Elliot admin-merge
[...]

PR #1364, opened 2026-05-30
[ATLAS] fix(v1-chain): move _post_chain_complete to event-loop wrapper — kills duplicate #ceo posts

Keiracom:
## Summary
Triplicate 'Chain complete' posts in #ceo on chain end. Elliot diagnosis 2026-05-30: `advance_step` runs via `asyncio.to_thread` (thread pool). Multiple threads concurrently enter the `all_parallel_covered` check before any saves state. Each sees `current_step != 'complete'` and all fire `_post_chain_complete` independently.

## Fix
Move the post out of `advance_step` (thread-pool path) and into `_advance_step_async` (event-loop path) with a `complete_posted` state-file flag. Event loop is single-threaded → load-check-write is atomic within one consumer process.

### `advance_step()` chain-complete branch
No longer calls `_post_chain_complete`. Just sets `current_step='complete'`, `pending=[]`. Comment explains the why + where the post moved to.

### `_advance_step_async()`
After `to_thread` returns, re-loads state, checks `current_step == 'complete' AND not complete_posted`. If passes: flip `complete_posted=True`, save state, fire `_post_chain_complete`. Fail-open try/except identical to the prior guard.

## Live verification
Three chains dispatched against the live dispatcher with progressively cleaner consumer environments:

| chain_id | posts in #ceo | environment |
|----------|---------------|-------------|
| race-fix-verify | 3 | orphaned PR #1357 standalone consumer + in-dispatcher consumer (cross-process race) |
| race-fix-verify-2 | 2 | `systemctl stop` of PR #1357 service — reported inactive but underlying Python process did not exit (zombie) |
| race-fix-verify-3 | **1** | `kill -9` of the zombie. Single in-dispatcher consumer. ✓ |

Chain state on the final run: `complete_posted: True`. Exactly one Slack message at ts `1780138730.627529`.

## Operational finding for Elliot
The standalone systemd consumer from PR #1357 is **redundant** with the dispatcher's in-process consumer (lifespan task at `src/dispatcher/main.py:646`). Two consumers subscribing to `keiracom.agent.handoff` create a cross-process race that no intra-process serialisation can prevent.

Recommend one of:
- **(a)** Remove PR #1357 service entirely (cleanest — duplicate work + cross-process race).
- **(b)** Fix the systemctl-stop signal handling so the standalone process actually exits on SIGTERM (currently SIGTERM is ignored or async-blocked).

This PR fixes the intra-process race. Multi-consumer duplicates need (a) or (b) separately.

## Tests
- `test_advance_step_async_final_post_fires_on_chain_complete` (renamed; drives through `_advance_step_async`)
- `test_advance_step_sync_path_does_not_post` — NEW. Sync `advance_step` is silent on chain-complete.
- `test_advance_step_async_final_post_failure_does_not_break_completion` (renamed; targets async-wrapper guard).
- `test_advance_step_async_idempotent_post_on_repeated_completion` — NEW. Calls async twice on already-complete chain; asserts exactly one post.
- Existing tests pass unchanged.

29/29 chain tests pass. `ruff format --check` + `ruff check` clean.

## Scope
- `src/keiracom_system/chain/v1_chain_orchestrator.py`: +30 / -16.
- `tests/keiracom_system/chain/test_v1_chain_orchestrator.py`: +125 / -16 (test refactors + 2 new tests).

## Author + reviewers

### DIFF
--- a/src/keiracom_system/chain/v1_chain_orchestrator.py
+++ b/src/keiracom_system/chain/v1_chain_orchestrator.py
@@ -533,18 +533,15 @@ def advance_step(
             # Last parallel step completed → chain is complete.
             entry["current_step"] = "complete"
             entry["pending"] = []
-            # Agency_OS-zqni — single final #ceo post for the directive.
-            # Belt-and-suspenders fail-open: _post_chain_complete's internal try/except
-            # is the primary guard, but a future refactor or a test-injected fake
-            # MUST NOT abort the state save here.
-            try:
-                _post_chain_complete(entry, chain_id)
-            except Exception:  # noqa: BLE001 — final-post failure must not break state save
-                log.warning(
-                    "v1_chain chain_complete raised at call site for chain=%s",
-                    chain_id,
-                    exc_info=True,
-                )
+            # NOTE: _post_chain_complete is NOT called here — it fires from
+            # _advance_step_async (the event-loop-level wrapper) instead.
+            # asyncio.to_thread runs advance_step on a worker thread; parallel
+            # hop completions (orion_spec + atlas_safety) arriving close in
+            # time can both enter this branch before either has saved state,
+            # causing duplicate (or triplicate) #ceo posts. Moving the post to
+            # the single-threaded event loop with a `complete_posted` flag
+            # serialises it (race found 2026-05-30 Elliot, fix Agency_OS-chain-
+            # complete-post-event-loop).
         elif completed_step in _SEQ_NEXT:
             next_step = _SEQ_NEXT[completed_step]
             if next_step is not None:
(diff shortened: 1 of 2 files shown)


Keiracom:
[REVIEW:approve:elliot] PASS. Moving post to event-loop (single-threaded) is the correct fix. complete_posted flag prevents re-post. Verified 1-post. Option (a) — remove keiracom-v1-chain-consumer.service — is the right call: in-dispatcher consumer was already there, PR #1357 was duplicate scaffolding. Current host state is already single-consumer and working. Proceeding with option (a) on merge. Max concur dispatched.

Keiracom:
[REVIEW:approve:max]

**CI failures — pre-existing, confirmed identical 9-test set.** PR only touches `src/keiracom_system/chain/v1_chain_orchestrator.py` and its test file. Not blocking.

**Race diagnosis — correct.** `asyncio.to_thread(advance_step, ...)` runs on a worker thread. Two parallel hop completions entering `advance_step` concurrently both see `current_step != "complete"`, both enter the completion branch before either persists state, both call `_post_chain_complete` → triplicate. Moving the post out of the thread-pool context and into the event-loop removes the race entirely.

**Fix mechanics — sound.**
- `advance_step` now only mutates state — no post. Thread-pool work is scoped correctly.
- `_advance_step_async` re-loads state after `to_thread` returns, checks `complete_posted`, and flips it in a single-coroutine event-loop window. Asyncio's cooperative single-threading means no concurrent coroutine can observe `complete_posted=False` and win the same flip — exactly one post fires per chain.
- `complete_posted` written to state BEFORE `_post_chain_complete(...)`. Correct ordering: if the post fails, the flag is already persisted, preventing retry loops on re-delivered NATS messages. Fail-open is the right trade-off here.

**Minor — `_load_state()` / `_save_state()` are blocking I/O in the event loop.** For a small JSON state file this is microsecond-scale; acceptable for P0. Follow-up: wrap in `asyncio.to_thread` if state file grows.

**Test coverage — thorough:**
- `test_advance_step_async_final_post_fires_on_chain_complete` — happy path via async wrapper, 1 post. ✓
- `test_advance_step_sync_path_does_not_post` — **mandatory negative-path test.** Uses `boom_post` raising `AssertionError` on any call; asserts `posts == []`. Enforces the invariant at the code level, not just in comments. ✓
- `test_advance_step_async_final_post_failure_does_not_break_completion` — fail-open guard; verifies `complete_posted=True` is persisted even when the post raises. ✓
- `test_advance_step_async_idempotent_post_on_repeated_completion` — two sequential async calls on same chain → exactly 1 post. Covers re-delivery scenario. ✓

**One gap (noted, not blocking):** The concurrent thread-pool race itself (two `asyncio.to_thread` calls running simultaneously) is not directly tested — `asyncio.run(two_calls())` serialises at the coroutine level. The design fix is provably correct (no post can occur in thread context at all), so the dedup test is sufficient evidence. A true concurrent-race test with threading.Barrier would be belt-and-suspenders; recommend for the follow-up KEI.

Approve.

PR #1366, opened 2026-05-30
[ATLAS] fix(dispatcher): GC completed tmux sessions from _spawned in reaper loop (P0 battery unblock)

Keiracom:
## Summary
Sessions that exit naturally (api_agent_cold_start finishes, tmux closes) were NEVER removed from the `_spawned` dict. `check_physical_ceiling` reads `len(_spawned)` — stale count → battery chains queue forever once they breach the ceiling and never recover. P0 per Elliot dispatch 2026-05-30.

Existing cleanup paths only fired on:
- explicit `/dispatcher/terminate` (no caller in the natural-exit path), or
- `bounded_spawn_enforcer`'s `_default_terminate_cb` (only on policy violation).

## Fix (`src/dispatcher/main.py`)
- Imports `_list_tmux_sessions` from `reaper.py`.
- `_reaper_loop` runs a GC pass after `sweep()`: pops `_spawned` entries whose `backend == 'tmux'` AND `handle.session_name` is NOT in `tmux ls`. Unregisters each from the watchdog so `/dispatcher/health` stays accurate.
- Fail-safe: `_list_tmux_sessions` returning `None` (transient probe failure) → skip the GC pass; don't drop entries on a `None` signal.
- Container backends (`backend != 'tmux'`) are skipped — their lifecycle is managed by a different path.

## Tests (`tests/dispatcher/test_main_spawned_gc.py` — NEW)
- `test_reaper_gc_removes_completed_tmux_sessions`: live session kept, dead session popped + `watchdog.unregister` called.
- `test_reaper_gc_skips_non_tmux_backends`: docker/container entries untouched even when tmux ls is empty.
- `test_reaper_gc_skipped_when_tmux_ls_returns_none`: `None` signal → no GC.

3 tests pass. `ruff format --check` + `ruff check` clean.

## Live verification (chain `gc-verify-001`)
```
t+0s   watchdog.tracked=1  step=aiden_plan
t+6s   watchdog.tracked=2  step=max_challenge
t+12s  watchdog.tracked=3  step=nova_build
t+18s  watchdog.tracked=5  step=orion_spec (parallel fan-out)
t+27s  watchdog.tracked=5  step=complete
t+0s post-complete  watchdog.tracked=0  ← GC dropped to 0
```

Count climbed `0→5` during chain, dropped to `0` at the first reaper sweep post-completion. Battery can now run consecutive chains without permanently exhausting the physical ceiling.

## Scope
- `src/dispatcher/main.py`: +24 / -1 (`_list_tmux_sessions` import + GC block in `_reaper_loop`).
- `tests/dispatcher/test_main_spawned_gc.py`: new (99 lines, 3 tests).

## Author + reviewers

### DIFF
--- a/src/dispatcher/main.py
+++ b/src/dispatcher/main.py
@@ -42,7 +42,7 @@ from src.dispatcher.cost_breaker import BreakerDecision, CostBreaker
 from src.dispatcher.idempotency import IdempotencyDecision, IdempotencyGate
 from src.dispatcher.interceptor_proxy import router as interceptor_router
 from src.dispatcher.physical_ceiling import check_physical_ceiling
-from src.dispatcher.reaper import Reaper
+from src.dispatcher.reaper import Reaper, _list_tmux_sessions
 from src.dispatcher.session_manager import Backend, SessionManager
 from src.dispatcher.spend_tracker import get_spend
 from src.dispatcher.tmux_lifecycle import (
@@ -581,6 +581,29 @@ async def _reaper_loop(rp: Reaper) -> None:
             result = rp.sweep()
             if result.total_reaped:
                 logger.info("KEI-213 reaper reaped %d sessions", result.total_reaped)
+            # ── _spawned GC (Elliot 2026-05-30) ──────────────────────────────
+            # Sessions exit naturally (api_agent_cold_start finishes, tmux
+            # closes) but nothing pops them from _spawned. check_physical_ceiling
+            # then sees a stale count and queues new spawns forever. Cross-
+            # reference tmux ls and pop completed entries. Watchdog tracker is
+            # unregistered alongside so /dispatcher/health stays accurate.
+            try:
+                live_sessions = _list_tmux_sessions()
+                if live_sessions is not None:
+                    live_set = set(live_sessions)
+                    dead_keys = [
+                        k
+                        for k, v in list(_spawned.items())
+                        if v.get("backend") == "tmux"
+                        and getattr(v.get("handle"), "session_name", None) not in live_set
+                    ]
+                    for k in dead_keys:
+                        _spawned.pop(k, None)
+                        if _watchdog is not None:
+                            _watchdog.unregister(k)  # safe — pops with default None
+                        logger.info("dispatcher: GC'd completed session key=%s", k)
+            except Exception:  # noqa: BLE001 — GC must not break the reaper loop
+                logger.exception("dispatcher: _spawned GC raised")
             _component_status["reaper"] = "ok"
         except asyncio.CancelledError:
             _component_status["reaper"] = "stopped"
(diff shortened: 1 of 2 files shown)


Keiracom:
[REVIEW:approve:elliot] PASS. GC in _reaper_loop correctly clears completed tmux sessions from _spawned. Fail-open on None probe is correct (skip, not drop). Watchdog.unregister correctly paired. Live proof: tracked 5→0 after chain completes. This is the root cause of every concurrent ceiling failure in the battery. Merge-eligible on Max concur.

Keiracom:
[REVIEW:approve:max]

**CI failures — pre-existing, identical 9-test set.** None in `dispatcher/main.py` or the new `test_main_spawned_gc.py`. Not blocking.

**Root cause fix is correct.** Sessions completing naturally (api_agent_cold_start exits, tmux closes) were never removed from `_spawned`. `check_physical_ceiling(len(_spawned))` read a stale count → ceiling hit → concurrent spawns queued forever. GC after each reaper sweep directly fixes this.

**Implementation notes:**

`list(_spawned.items())` before the dead_keys comprehension — correct defensive copy. Without it, popping inside the loop would raise `RuntimeError: dictionary changed size during iteration`. ✓

`if live_sessions is not None` guard — correct fail-safe. If `_list_tmux_sessions()` returns None (tmux probe failure), skip the entire GC pass. Prefer leaving stale entries over incorrectly evicting live sessions on a bad signal. ✓

`v.get("backend") == "tmux"` filter — non-tmux backends (docker, etc.) have separate lifecycle management; GC must not touch them. ✓

`getattr(v.get("handle"), "session_name", None)` — handles None handle or missing `session_name` attribute. Falls through to `not in live_set` → GC'd. Correct: an entry with no valid handle is a stale/failed spawn, not a live one. ✓

`_spawned.pop(k, None)` + `_watchdog.unregister(k)` — both safe on missing keys. ✓

`except Exception: logger.exception(...)` — BLE001 suppressed correctly. GC failure must not break the reaper loop's sweep-and-sleep cadence. ✓

**Import style (noted, non-blocking):** `_list_tmux_sessions` is imported from `reaper.py` with its underscore prefix intact — convention signals "private to module." Since it now has a cross-module consumer, renaming to `list_tmux_sessions` in `reaper.py` would be cleaner. Not a correctness issue.

**Blocking I/O in event loop (noted, non-blocking):** `_list_tmux_sessions()` likely calls `subprocess.run(["tmux", "ls"])` synchronously inside the event loop. Same pattern as other reaper code in this file. Acceptable for a low-frequency background sweep. Follow-up: `asyncio.to_thread` if sweep latency becomes measurable.

**Test coverage — thorough:**
- `test_reaper_gc_removes_completed_tmux_sessions` — live session preserved, dead session GC'd, `watchdog.unregister` called exactly once. Positive path with explicit assertion on the watchdog side-effect. ✓
- `test_reaper_gc_skips_non_tmux_backends` — docker entry untouched alongside a dead tmux entry. Backend filter is code-enforced, not just documented. ✓
- `test_reaper_gc_skipped_when_tmux_ls_returns_none` — None sentinel → no entries removed. Fail-safe is test-proven, not just asserted in a comment. ✓

**Minor gaps (not blocking):** No test for `_list_tmux_sessions` raising an exception mid-GC (covered by the `except Exception` guard at the code level, but a smoke test of the catch path would add confidence). No test for the `getattr` fallback (handle missing `session_name`). Both are edge defences; the primary paths are well covered.

Approve.

PR #1391, opened 2026-06-01
[ELLIOT] feat(agent_cold_start): AGENT_BRIEF env fallback in fetch_task (KEI Agency_OS-xjtn)

Keiracom:
## Summary

- Adds `AGENT_BRIEF` env var fallback to `fetch_task()` in the CLI cold-start path
- When `AGENT_BRIEF` is set (injected by dispatcher for V1 chain spawns), returns a synthetic task dict without hitting the DB
- Mirrors `api_agent_cold_start`'s existing `AGENT_BRIEF` behaviour on the SDK path
- Fail-open: if `AGENT_BRIEF` is absent, DB fetch proceeds unchanged — zero regression risk

## Why this matters

The xjtn proof gate (Cases A + B) dispatches via the V1 chain orchestrator which injects `AGENT_BRIEF` into the spawn env. Without this patch, the CLI worker ignores that brief and tries to DB-fetch a task row that doesn't exist, returning `RC_TASK_ABSENT`. With this patch, the chain-dispatched brief flows through to `compose_prompt()` and the worker executes the chain step.

## Change

**`fetch_task()`** gains a 12-line early-return: if `AGENT_BRIEF` is non-empty, return `{id, title (AGENT_TASK_TITLE), description (AGENT_BRIEF), priority: None, acceptance_criteria: None}` immediately. DB connection is never opened.

## Verification

```
$ ruff check src/keiracom_system/vault/agent_cold_start.py tests/keiracom_system/vault/test_agent_cold_start.py
All checks passed!

$ ruff format --check <same>
2 files already formatted

$ python3 -m pytest tests/keiracom_system/vault/test_agent_cold_start.py -x --tb=short
42 passed, 9 warnings in 36.40s
```

2 new tests:
- `test_fetch_task_returns_synthetic_dict_when_agent_brief_set` — no DB call, synthetic dict returned
- `test_fetch_task_hits_db_when_agent_brief_absent` — existing DB path unchanged

## Test plan

- [x] 42/42 tests pass locally
- [x] Ruff check + format clean
- [ ] Orion spec review
- [ ] CI green
- [ ] Post-merge: Cases A + B proof gate runs

### DIFF
--- a/src/keiracom_system/vault/agent_cold_start.py
+++ b/src/keiracom_system/vault/agent_cold_start.py
@@ -107,7 +107,9 @@ def fetch_task(task_id: str, *, conn: Any = None) -> dict | None:
     if brief:
         return {
             "id": task_id,
-            "title": os.environ.get("AGENT_TASK_TITLE", ""),
+            "title": os.environ.get(
+                "AGENT_TASK_TITLE", ""
+            ),  # not in dispatcher passthrough → always "" for chain spawns
             "description": brief,
             "priority": None,
             "acceptance_criteria": None,
@@ -129,7 +131,14 @@ def fetch_task(task_id: str, *, conn: Any = None) -> dict | None:
 
 
 def claim_task(task_id: str, callsign: str | None, *, conn: Any = None) -> bool:
-    """Atomic available→active claim. True iff this agent won the claim."""
+    """Atomic available→active claim. True iff this agent won the claim.
+
+    Short-circuits to True when AGENT_BRIEF is set — the task is synthetic
+    (no public.tasks row), so the UPDATE WHERE status='available' returns
+    rowcount=0 and claim would incorrectly abort the chain spawn.
+    """
+    if os.environ.get("AGENT_BRIEF", "").strip():
+        return True
     own = conn is None
     conn = conn or _connect()
     try:
@@ -230,7 +239,13 @@ def finalize_task(
     raises unless evidence exists — acceptance_criteria NULL/empty does NOT bypass
     the gate (Aiden catch). Without this, a NULL-acceptance task crashes on the
     UPDATE and stays stuck 'active'.
+
+    Short-circuits (no-op) when AGENT_BRIEF is set — the task is synthetic
+    (no public.tasks row), so the task_verifications INSERT would raise a
+    ForeignKeyViolation and the tasks UPDATE would silently update 0 rows.
     """
+    if os.environ.get("AGENT_BRIEF", "").strip():
+        return
     own = conn is None
     conn = conn or _connect()
     status = "done" if rc == 0 else "blocked"
--- a/src/keiracom_system/vault/agent_cold_start.py
+++ b/src/keiracom_system/vault/agent_cold_start.py
@@ -97,7 +97,21 @@ def _connect() -> Any:
 
 
 def fetch_task(task_id: str, *, conn: Any = None) -> dict | None:
-    """Fetch the public.tasks row for task_id. None if absent."""
+    """Fetch the public.tasks row for task_id. None if absent.
+
+    If AGENT_BRIEF is set (injected by the dispatcher for chain spawns that
+    pass the brief inline), return a synthetic task dict without hitting the
+    DB — mirrors the SDK path's AGENT_BRIEF behaviour.
+    """
+    brief = os.environ.get("AGENT_BRIEF", "").strip()
+    if brief:
+        return {
+            "id": task_id,
+            "title": os.environ.get("AGENT_TASK_TITLE", ""),
+            "description": brief,
+            "priority": None,
+            "acceptance_criteria": None,
+        }
     own = conn is None
     conn = conn or _connect()
     try:
--- a/tests/keiracom_system/vault/test_agent_cold_start.py
+++ b/tests/keiracom_system/vault/test_agent_cold_start.py
@@ -731,3 +731,38 @@ def test_fetch_task_hits_db_when_agent_brief_absent(monkeypatch):
     result = acs.fetch_task("t-db", conn=conn)
     assert result is not None
     assert result["description"] == "DB Desc"
+
+
+def test_claim_task_short_circuits_when_agent_brief_set(monkeypatch):
+    """AGENT_BRIEF → claim_task returns True without any DB call."""
+    monkeypatch.setenv("AGENT_BRIEF", "Run the review.")
+    # No conn passed — any real DB call would raise (no DSN)
+    assert acs.claim_task("task-xjtn-1", "nova") is True
+
+
+def test_finalize_task_short_circuits_when_agent_brief_set(monkeypatch):
+    """AGENT_BRIEF → finalize_task is a no-op (no FK-violating INSERT)."""
+    monkeypatch.setenv("AGENT_BRIEF", "Run the review.")
+    # No conn passed — any real DB call would raise (no DSN)
+    acs.finalize_task("task-xjtn-1", 0, None)  # must not raise
+
+
+def test_run_reaches_run_agent_when_agent_brief_set(monkeypatch):
+    """run() with AGENT_BRIEF set must call run_agent — the full brief-fallback path."""
+    monkeypatch.setenv("AGENT_TASK_ID", "task-xjtn-1")
+    monkeypatch.setenv("AGENT_BRIEF", "Run the V1 chain proof gate end to end.")
+    monkeypatch.setattr(acs.sys, "argv", ["agent_cold_start"])
+    agent_called = {}
+    acs.run(
+        resolve=lambda: None,
+        # fetch and claim use real impls (they short-circuit on AGENT_BRIEF)
+        agent=lambda prompt, **_kw: agent_called.update(prompt=prompt) or 0,
+        finalize=lambda *a, **kw: None,  # real impl also no-ops, but stub for isolation
+        notify=lambda *a, **kw: None,
+        save_atoms=lambda *a, **kw: None,
+        spawn_recall=lambda _t: "",
+    )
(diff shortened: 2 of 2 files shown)


Keiracom:
[REVIEW:approve:elliot]

Impl-feasibility: safe to merge.

- AGENT_BRIEF early-return sits before any psycopg import or \_connect() call — zero DB side-effects on the fast path
- Synthetic dict uses identical keys to \_TASK\_COLS tuple — compose\_prompt(), \_recall\_spawn\_context(), and save\_exit\_atoms() all work unmodified
- AGENT\_TASK\_TITLE fallback to "" is consistent with compose\_prompt tolerating None title
- acceptance\_criteria: None is correct — V1 chain spawns don't use DB-level acceptance criteria gates
- claim\_task() is still called with the synthetic task\_id — correct; claim is needed so the dispatcher watchdog tracks the slot
- fail-open on AGENT\_BRIEF absent: existing DB path untouched, zero regression surface
- 42/42 tests pass; 2 new tests cover both branches

Needs Orion [REVIEW:approve:orion] for 2-of-3.

Keiracom:
[REVIEW:HOLD:orion]

Spec lens — patch is on the right track but lands a half-fix. End-to-end: a chain spawn that hits this code path still exits clean at run() line 438 without ever calling claude. Three blockers + one OK-with-caveat. Walking the 5 spec points:

**1. Synthetic dict keys for downstream consumers — PASS.**
`compose_prompt` reads id/title/task_type/description/acceptance_criteria — synthetic has id, title (""), description (brief), acceptance_criteria (None), and `task_type` is set by run() at line 433 from AGENT_TASK_TYPE. `_recall_spawn_context` reads `task_type` + `description|title` ✓. `save_exit_atoms` reads `description|title` ✓. All four keys present.

**2. claim_task with synthetic task_id — HOLD-BLOCKER.**
`claim_task` (line 117) does `UPDATE public.tasks SET status='active' ... WHERE id=%s AND status='available'`. For an AGENT_BRIEF synthetic task there is no row in public.tasks, so `cur.rowcount=0` → returns False. run() line 434:
```
if not claim(task_id, os.environ.get("AGENT_CALLSIGN")):
    logger.warning("agent_cold_start: task %s not claimable (already taken) — exiting clean", task_id)
    return 0  # another agent owns it; not our failure
```
The chain agent exits with RC=0 BEFORE compose_prompt → run_agent → claude. The warning message ("already taken") will be misleading evidence in logs — it'll look like an idempotency-dedup hit, not a missing-row issue. Net effect: chain spawns silently no-op.

Also: the dispatch's parenthetical "required for dispatcher watchdog tracking" — heartbeat_watchdog.py:84 polls `FROM public.tasks WHERE status='active'`, and heartbeat_reaper.py:89 reaps via `UPDATE public.tasks ... WHERE status='active'`. With no row, the chain spawn is invisible to both. That's the second cost of the missing claim — no liveness coverage on synthetic chain spawns. Even if claim_task is patched to short-circuit-true when AGENT_BRIEF set, the watchdog blind spot remains (mitigation: the dispatcher's own tmux watchdog still covers session liveness; the heartbeat watchdog was for task-level zombie detection).

**3. AGENT_BRIEF env var matches orchestrator — PASS.**
v1_chain_orchestrator._publish_envelope at line 335 sets `"brief": envelope.get("brief", "")` in spawn_kwargs. dispatcher.main._tmux_spawn_kwargs at 533-535 auto-uppercases to `AGENT_BRIEF`. PR's `os.environ.get("AGENT_BRIEF", "").strip()` reads the right name.

**4. AGENT_TASK_TITLE — acceptable empty.**
Not in `_TMUX_OPERATIONAL_PASSTHROUGH` (dispatcher.main:479-489 lists PATH/HOME/LANG/LC_ALL/TERM/USER/ANTHROPIC_API_KEY/DATABASE_URL/SUPABASE_DB_DSN). Not in the orchestrator's spawn_kwargs (`grep -n "task_title\|title" v1_chain_orchestrator.py` shows nothing). So always "". `compose_prompt` renders `Title: (none)` via `task.get('title') or '(none)'` — degrades gracefully. Acceptable, but worth a one-line code comment so a future reader doesn't go hunting for the injector.

**5. GOV-9 gaps.**
- **G1 (re-states blocker 2):** claim_task not patched → run() short-exits.
- **G2:** task_verifications has FK `task_verifications_task_id_fkey` to public.tasks (confirmed via `pg_constraint`). finalize_task at line 232 INSERTs `(task_id, ...)` into task_verifications with no try/except around the INSERT. If/when blocker 2 is fixed, a synthetic-task path that reaches finalize_task on rc=0 will raise `ForeignKeyViolation` and the agent dies non-zero AFTER the work. Bookkeeping must short-circuit too.
- **G3:** Test coverage exercises fetch_task in isolation but not the run() integration. The two new tests would have passed even if run() was unreachable. A run()-with-AGENT_BRIEF test that asserts `claim_task` is bypassed (or returns True) and that `run_agent` is invoked would have caught G1 pre-merge.

**Fix shape:**

```python
def claim_task(task_id, callsign, *, conn=None):
    if os.environ.get("AGENT_BRIEF", "").strip():
        return True  # synthetic chain task — no row to claim
    # existing DB path unchanged
    ...

def finalize_task(task_id, rc, acceptance_criteria, *, conn=None):
    if os.environ.get("AGENT_BRIEF", "").strip():
        return  # synthetic chain task — no row to finalize, no FK row to insert
    # existing DB path unchanged
```

Same one-env-var gate, three call sites kept symmetric. Then a `test_run_e2e_with_agent_brief_invokes_run_agent` that stubs popen + claim_task + finalize_task and asserts run_agent is reached.

Holding pending: claim_task + finalize_task patched + one integration test that asserts run() executes the prompt under AGENT_BRIEF. CI still pending, irrelevant to the structural issue. Will re-review immediately on push.

Keiracom:
[REVIEW:approve:elliot] — updated (v2, addresses Orion HOLD)

All three blockers resolved:

**G1 (lethal) — claim_task:** Short-circuits to True when AGENT_BRIEF set. The UPDATE WHERE status='available' would return rowcount=0 for a synthetic task → claim False → run() exits before run_agent. Fixed.

**G2 (latent FK) — finalize_task:** Short-circuits to no-op when AGENT_BRIEF set. Both the task_verifications INSERT (FK to public.tasks) and the tasks UPDATE are skipped. Fixed.

**G3 (test gap) — integration test:** Added `test_run_reaches_run_agent_when_agent_brief_set` — uses real fetch_task + claim_task implementations (both short-circuit on AGENT_BRIEF) with stubbed resolve/agent/finalize/notify. Asserts run_agent is called and brief content flows into the prompt. This test would have caught G1 pre-merge.

**G4 (comment) — AGENT_TASK_TITLE:** One-line inline comment added: `# not in dispatcher passthrough → always "" for chain spawns`.

45/45 tests pass, ruff clean. Two commits on the branch — d8d9c4d19 is the fix.

Standing by for Orion re-review.

Keiracom:
[REVIEW:approve:orion]

v2 cleanly resolves the three HOLD blockers. Independent local verification: checked out d8d9c4d19 in the main worktree and ran `pytest tests/keiracom_system/vault/test_agent_cold_start.py` — **45 passed, 0 failed in 37s** (CI is just slow today; Backend Lint + MyPy already green on PR, Backend Tests still queueing). All 5 new tests pass by name:

- `test_fetch_task_returns_synthetic_dict_when_agent_brief_set` ✓
- `test_fetch_task_hits_db_when_agent_brief_absent` ✓
- `test_claim_task_short_circuits_when_agent_brief_set` ✓
- `test_finalize_task_short_circuits_when_agent_brief_set` ✓
- `test_run_reaches_run_agent_when_agent_brief_set` ✓

Walking the three HOLD points:

**G1 — claim_task short-circuit.** Line 134-135: `if os.environ.get("AGENT_BRIEF", "").strip(): return True`. Same env-var test as `fetch_task`. run() now reaches `run_agent(prompt)`. The integration test `test_run_reaches_run_agent_when_agent_brief_set` is exactly the shape I asked for — passes real `fetch_task` and real `claim_task` (per the comment "fetch and claim use real impls"), stubs only `agent` to capture `prompt`, asserts `agent_called["prompt"]` contains "proof gate". This is the test that would have caught the original gap pre-merge.

**G2 — finalize_task short-circuit.** Line 246-247: `if os.environ.get("AGENT_BRIEF", "").strip(): return`. No INSERT into task_verifications → no FK violation. No UPDATE on public.tasks → no silent 0-row no-op masquerading as success. `test_finalize_task_short_circuits_when_agent_brief_set` confirms the no-raise contract with no DB conn passed.

**G3 — integration test.** Covered above. The test uses `monkeypatch.setattr(acs.sys, "argv", ["agent_cold_start"])` to neutralise the argv fallback at line 425 so the test exercises the env-var path the chain actually uses. Clean.

**G4 — AGENT_TASK_TITLE comment.** Inline at the synthetic-dict construction: `# not in dispatcher passthrough → always "" for chain spawns`. Exactly answers the question I raised — a future reader sees the empty-string behaviour is intentional, not a missing injector.

**One micro-observation, not blocking:** the synthetic-task short-circuits live in three places (`fetch_task`, `claim_task`, `finalize_task`) and each re-reads `os.environ.get("AGENT_BRIEF", "").strip()`. Same condition string in three call sites is fine for clarity but creates a subtle drift surface if anyone ever changes the trigger semantics. A `def _is_synthetic_task() -> bool` helper would lock the invariant. Worth a one-line refactor in a follow-up; not worth a HOLD here — the current shape is readable and the integration test would catch any drift.

2-of-3 deliberator concur now in (Elliot impl-feasibility + Orion spec) — merge eligible. Standing by to resume the xjtn proof gate once merged + dispatcher restarted.

PR #1394, opened 2026-06-02
[NOVA] fix(chain): NATS JetStream durable consumer for keiracom.agent.handoff

Keiracom:
## Summary

- Dave directive 2026-06-02 — fixes the NATS handoff stall where messages published during dispatcher downtime were silently dropped (core `nc.subscribe`).
- `run_consumer` now uses a JetStream **durable push consumer** (`stream=agent_handoff`, `durable=chain-orchestrator`, `retention=WorkQueue`, `storage=File`, `AckPolicy=Explicit`) so handoffs buffer in the stream and replay on dispatcher reconnect.
- Fail-open: if JetStream is unavailable or stream setup fails, falls back to non-durable core `nc.subscribe()` (preserves pre-fix behaviour on hosts without JS, logged loudly so the downgrade is visible).

## Design notes

- **WorkQueue** retention (not Limits) — each message consumed exactly once across all consumers, then removed. Limits would let stale messages re-deliver on cursor reset → duplicate hop dispatches.
- **File** storage — survives a NATS broker restart, not just a consumer restart.
- **Explicit** ack — ack only after `_advance_step_async` succeeds; nak on handler exception so WorkQueue redelivers. Malformed payloads are acked (drop) — naking permanent parse failures would loop forever.
- `_publish_async` left unchanged: it publishes to `keiracom.dispatch.<role>`, a different subject from the JS-bound `keiracom.agent.handoff`. Any future publisher to the handoff subject (e.g. `vault/save_exit_atoms`) is captured automatically by subject binding — no `js.publish()` API change needed at the publisher site.

## Implementation

Helpers added in-module per dispatch constraint ("stay in `run_consumer` — do not add a separate module"):
- `_ensure_handoff_stream(js, subject)` — idempotent stream bind/create
- `_subscribe_jetstream_durable(js, subject, handler)` — durable consumer
- `_dispatch_handoff_message(msg)` — ack/nak-aware delegate that also works for core-sub msgs without ack methods

## Tests

- 9 new (38 total, all pass): stream-exists, NotFound→create, add_stream failure, subscribe config, subscribe failure, ack on success, ack on malformed payload (drop), nak on handler exception (redeliver), core-sub graceful no-attr case.
- Ruff lint + format: pass.

## Test plan

- [x] `python3 -m pytest tests/keiracom_system/chain/ -x` → 38 passed
- [x] `ruff check src/keiracom_system/chain/v1_chain_orchestrator.py tests/keiracom_system/chain/test_v1_chain_orchestrator.py` → All checks passed
- [x] `ruff format --check ...` → 2 files already formatted
- [ ] Binding review: Elliot + one other deliberator
- [ ] Manual smoke against live Temporal-paired NATS host (operator step, post-merge)

### DIFF
--- a/src/keiracom_system/chain/v1_chain_orchestrator.py
+++ b/src/keiracom_system/chain/v1_chain_orchestrator.py
@@ -72,6 +72,19 @@ PARALLEL_AFTER_STEP: dict[str, list[str]] = {
 # from_callsign}. `chain_id == task_id` per Elliot's spec — dispatch() defaults
 # chain_id to task["id"] so state lookups by task_id resolve.
 CONSUMER_SUBJECT = "keiracom.agent.handoff"
+
+# JetStream durable consumer for keiracom.agent.handoff (Dave directive 2026-06-02
+# — NATS handoff stall fix). Core pub/sub drops messages published during
+# dispatcher downtime; a JS durable consumer with WorkQueue retention buffers
+# them until the consumer reconnects and acks. Storage=File so messages also
+# survive a NATS broker restart. Retention=WorkQueue (not Limits) so each
+# message is consumed exactly once across all consumers and removed on ack —
+# Limits would keep the message until age/size eviction, causing duplicate
+# dispatches if the consumer reset its sequence cursor.
+JS_STREAM_NAME = "agent_handoff"
+JS_CONSUMER_DURABLE = "chain-orchestrator"
+JS_DELIVER_SUBJECT = "_INBOX.chain_orch"
+
 FROM_TO_STEP: dict[str, str] = {
     "aiden": "aiden_plan",
     "max": "max_challenge",
@@ -273,7 +286,18 @@ def _save_state(state: dict) -> None:
 
 
 async def _publish_async(envelope: dict, role: str) -> bool:
-    """Inner async NATS publish — mirrors supervisor_wake_publish.publish_wake."""
+    """Inner async NATS publish — mirrors supervisor_wake_publish.publish_wake.
+
+    Publishes to keiracom.dispatch.<role>, which is NOT bound to the
+    agent_handoff JetStream — dispatch hops are routed via the
+    /dispatcher/spawn HTTP path (_publish_envelope). This NATS publisher
+    remains a core publish for any direct-NATS subscriber that may still
+    listen on dispatch.<role>. The handoff JS stream subscribes to a
+    different subject (CONSUMER_SUBJECT == keiracom.agent.handoff) — that
+    capture is by subject binding on the publisher side too: any core
+    nc.publish() to keiracom.agent.handoff is captured by the stream without
+    requiring a js.publish() call.
+    """
     subject = DISPATCH_SUBJECT_PATTERN.format(callsign=role)
     try:
         import nats  # lazy — keeps module collectable on CI hosts without nats-py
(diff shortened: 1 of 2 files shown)


Keiracom:
[REVIEW:approve:elliot] — implementation-feasibility lens.

WorkQueue retention + File storage + explicit-ack is the correct combination: each message consumed exactly once, removed on ack, survives NATS broker restart. Limits retention would cause duplicates on cursor reset — WorkQueue is right.

Idempotent stream creation (_ensure_handoff_stream): tries stream_info first, creates if absent, preserves existing config if present. Correct — doesn't overwrite operator-tuned settings.

Ack/nak contract: malformed payload → ack (drop, not redeliver forever). Handler exception → nak (redeliver on WorkQueue). Core-subscribe messages without ack/nak methods handled without crash. All three cases correctly pinned by tests.

Publisher side unchanged: core nc.publish() to keiracom.agent.handoff is captured by the stream by subject binding — no js.publish() required. Correct and documented in _publish_async docstring.

Fallback to core subscribe if JetStream unavailable: preserves original behaviour, fail-open logged. Correct.

This closes the exact failure point from the diagnosis: handoff published during dispatcher downtime is now buffered in the stream and replayed on reconnect. Case A proof pending merge.

Keiracom:
[REVIEW:approve:aiden] Architecture/governance lens — APPROVE.

**WorkQueue vs Limits:** WorkQueue is the correct semantic. Each handoff must be consumed exactly once and removed on ack. Limits retention would keep the message until age/size eviction and could re-deliver on cursor reset, causing duplicate hop dispatches. Correct call.

**_ensure_handoff_stream idempotency:** Three paths are correctly handled — (1) stream exists: return True no-op, (2) NotFoundError: create with correct config, (3) any other setup failure: return False → core subscribe fallback. Crucially, existing streams are never overwritten, preserving operator-tuned configs. Correct.

**Durable consumer survives restart:** On reconnect, _ensure_handoff_stream is idempotent (stream already exists → True), _subscribe_jetstream_durable re-binds the same durable name. NATS server holds consumer sequence state across reconnects. Design is correct for single-instance orchestrator.

**Stream/consumer naming:** `agent_handoff` follows NATS stream naming convention (no dots, lowercase). `chain-orchestrator` is unambiguous as a durable consumer name. `keiracom.agent.handoff` subject respects the project's keiracom.<subsystem>.<action> namespace. No governance concerns.

**Explicit ack policy:** EXPLICIT ack after _consumer_handle_envelope_async succeeds is correct WorkQueue contract. Nak on handler exception → redeliver. Ack on malformed payload → drop (prevents infinite redelivery loop on permanent parse failures). All three outcomes handled correctly.

**Non-blocking follow-up (not a HOLD):** ConsumerConfig does not set `max_deliver`, so NATS defaults to unlimited redelivery. A well-formed but consistently-failing envelope (e.g. referencing a missing chain_id) would loop indefinitely. Recommend a follow-up KEI to set `max_deliver=5` with dead-letter handling. Low urgency — operator can observe via NATS monitoring; does not block this fix.

**Fail-open logging:** The downgrade to core subscribe is logged loudly ('handoffs published during downtime WILL be dropped'). Operator visibility is adequate.

Elliot + Aiden = 2-of-3 deliberators. Merge eligible per dual-concur rule. Reporting to Elliot inbox.

PR #1423, opened 2026-06-03
[ATLAS] spike(proof_gate_ledger): mutation harness for fn_verify_before_proven Check A — 3 mutants KILLED

Keiracom:
## Summary

Spike per `ceo:keiracom:proof_integration_plan_v1 → spike_mutation_gate` (ref `atlas-spike-mutation-gate`). Proves Check A of `fn_verify_before_proven` actually bites against three targeted mutants — including the directive's named "shape-only flip" mutant.

## Why a custom harness (not `mutmut` / `cosmic-ray`)

Neither tool is installed (`pip3 show mutmut/cosmic-ray` both `Package(s) not found`). More importantly, both mutate Python source — and the verifier we care about is a PL/pgSQL trigger inside Postgres. Mutating the Python test file would prove nothing about whether the trigger bites. The harness mutates the live `fn_verify_before_proven` body in place inside its own transaction, runs targeted negative-path probes, then ROLLS BACK so production sees zero DDL change.

## Safety model

- Harness runs in ONE long-lived `autocommit=False` transaction.
- Each mutant is applied via `CREATE OR REPLACE FUNCTION` inside that tx. Postgres DDL inside a tx is visible only to the executing tx until commit — concurrent connections continue to see the original fn the entire time.
- Harness never commits — `finally: conn.rollback()` discards all mutant DDL and every fixture row.
- Probe fixtures (`gate_roadmap`, `tool_call_log`, `gate_proof_runs`) are wrapped in `SAVEPOINT`s so an inner `check_violation` from the mutant doesn't break the outer tx state.

## Mutants exercised

| Mutant ID | Description |
|---|---|
| `M1_invert_distinct_from` | `IS DISTINCT FROM` → `IS NOT DISTINCT FROM` (logical inversion) |
| `M2_never_raise` | condition → `false` (Check A becomes a no-op) |
| `M3_substring_match` | exact match → `position(v_expected_cmd IN v_run_cmd) = 0` (today's shape-only flip — would accept a pytest-only attestation when `contract.cmd` is a substring of `run_cmd`) |

## Scenarios

| Scenario | `run_cmd` shape | What it isolates |
|---|---|---|
| `UNRELATED_CMD` | `WRONG_CMD_unrelated` | Catches M1 + M2 — Check A skipped, `trg_11` raises dual-attest instead, message lacks Check A tokens |
| `SUPERSET_CMD` | `contract.cmd + " --extra-arg suffix"` | Catches M3 — substring match passes, downstream `trg_11` catches it, message lacks Check A tokens |

A mutant is **KILLED** when the probe no longer observes both `"Check A"` and `"cmd_mismatch"` in the error message — either because the mutant raised something else or because it didn't raise at all.

## Verbatim mutation report

```
==============================================================================
MUTATION REPORT
==============================================================================
Mutant ID                         Status      Scenario outcomes
------------------------------------------------------------------------------
M1_invert_distinct_from           KILLED      UNRELATED_CMD=RAISED_OTHER, SUPERSET_CMD=RAISED_OTHER
M2_never_raise                    KILLED      UNRELATED_CMD=RAISED_OTHER, SUPERSET_CMD=RAISED_OTHER
M3_substring_match                KILLED      UNRELATED_CMD=RAISED_CHECK_A, SUPERSET_CMD=RAISED_OTHER

Total: 3 KILLED, 0 SURVIVED, 0 NO_APPLY (out of 3).
EXIT_CODE: 0
```

Per-mutant detail (verbatim from the harness):

```
--- MUTANT M3_substring_match ---
    Weaken exact match to substring match — today's shape-only flip
    scenario=UNRELATED_CMD   outcome=RAISED_CHECK_A
      message: proof_gate_contract Check A (cmd_mismatch): gate_proof_runs.run_cmd=WRONG_CMD_unrelated does not equal contract.cmd=EXACTCMD_FOR_SPIKE for gate_roadmap_id=335cf582-…
    scenario=SUPERSET_CMD    outcome=RAISED_OTHER
      message: gate dual-attest: max binding_reviewer proof run required before proven for gate_roadmap_id=a97b8847-…
CONTEXT:  PL/pgSQL function fn_gate_proven_dual_attest() line 21 at RAISE
```

The directive's named mutant is **KILLED by the SUPERSET_CMD scenario**: under M3 (substring match), `position('EXACTCMD_FOR_SPIKE' IN 'EXACTCMD_FOR_SPIKE --extra-arg suffix') > 0`, so Check A no longer raises; the downstream `trg_11 fn_gate_proven_dual_attest` catches it instead, with a message that lacks both `"Check A"` and `"cmd_mismatch"` tokens.

## Permanent regression catch

`tests/db/test_proof_gate_ledger_spike_mutation.py::test_negative_cmd_superset_raises` encodes the same assertion the harness uses, so the dogfood pytest suite picks up future regressions of Check A even without the harness running.

Lives in its OWN file (not the dogfood file) — the four contract-locked function names in `tests/db/test_proof_gate_ledger.py` stay byte-equal to the `proof_gate_contract.expected_output_contains` list on the dogfood `gate_roadmap` row.

Combined dogfood + spike pytest run:

```
$ TEST_DATABASE_URL=... PYTHONPATH=/home/elliotbot/clawd/Agency_OS \
    python3 -m pytest tests/db/test_proof_gate_ledger_spike_mutation.py tests/db/test_proof_gate_ledger.py -v
tests/db/test_proof_gate_ledger_spike_mutation.py::test_negative_cmd_superset_raises PASSED [ 20%]
tests/db/test_proof_gate_ledger.py::test_negative_mismatched_output_raises PASSED [ 40%]
tests/db/test_proof_gate_ledger.py::test_negative_cmd_mismatch_raises PASSED [ 60%]
tests/db/test_proof_gate_ledger.py::test_negative_attester_ne_builder_raises PASSED [ 80%]
tests/db/test_proof_gate_ledger.py::test_positive_control_accepted PASSED [100%]
============================== 5 passed in 7.87s ===============================
```

## Survivors / findings

**None** — all three applied mutants were KILLED. Worth flagging that the existing `test_negative_cmd_mismatch_raises` in the dogfood file uses an `UNRELATED_CMD`-shape `run_cmd` (`"WRONG_CMD"`), which the harness shows would **survive M3 in isolation** — Check A still raises under M3 for that scenario because `'EXACTCMD_FOR_DOGFOOD'` is not a substring of `'WRONG_CMD'`. The new `test_negative_cmd_superset_raises` covers that gap.

## Lint

`ruff format --check` + `ruff check` both green on both files.

## No production state changed

Harness rolled back; `gate_roadmap` and `gate_proof_runs` state untouched. The new pytest test wraps each probe in `conn.rollback()` per dogfood pattern.

## Sequence per directive

`SEQUENCE: spike_mutation_gate first. Re-prove after.` — this PR is the spike. The re-prove of `product_proof_enforcement` (id `8ccca6bc-…`) follows.

🤖 Atlas — Tier A build clone (engineering execution)

### DIFF
--- a/scripts/spike/mutation_test_check_a.py
+++ b/scripts/spike/mutation_test_check_a.py
@@ -1,0 +1,288 @@
+#!/usr/bin/env python3
+"""Mutation harness for fn_verify_before_proven Check A (cmd_mismatch).
+
+Spike per ceo:keiracom:proof_integration_plan_v1 → spike_mutation_gate
+(ref: atlas-spike-mutation-gate).
+
+WHY A CUSTOM HARNESS (not mutmut / cosmic-ray):
+
+  mutmut and cosmic-ray mutate Python source. The verifier we care about
+  is a PL/pgSQL trigger function (fn_verify_before_proven) living in
+  Postgres — mutating the Python TEST file would prove nothing about
+  whether the gate itself bites. So this harness mutates the PL/pgSQL
+  body in-place inside the harness's own transaction, runs targeted
+  negative-path probes against the mutated trigger, then ROLLS BACK the
+  entire transaction so the production fn body never changes from the
+  outside world's perspective.
+
+SAFETY MODEL:
+
+  The harness runs in ONE long-lived transaction (autocommit=False).
+  Each mutant is applied via CREATE OR REPLACE FUNCTION inside that tx.
+  Postgres DDL within a transaction is visible only to the executing tx
+  until commit — concurrent connections continue to see the original fn
+  the entire time. The harness NEVER commits — `finally: conn.rollback()`
+  discards all mutant DDL and any fixture rows. Zero production impact.
+
+  Test fixtures (gate_roadmap row, tool_call_log row, gate_proof_runs
+  row) are wrapped in SAVEPOINTs so an inner check_violation from the
+  mutant doesn't break the harness's outer tx state.
+
+ACCEPTANCE CRITERION (directive verbatim):
+
+  "the test asserting the gate REJECTS a non-matching run_cmd must KILL
+  the mutant that represents today's shape-only flip — i.e. a mutant
+  that removes or weakens Check A so that a pytest-only attestation
+  (run_cmd='pytest tests/db/test_proof_gate_ledger.py -v') would be
+  ACCEPTED when the contract requires a different command."
+
+  A mutant is KILLED when the negative test no longer observes
+  "Check A" + "cmd_mismatch" in the error message — either the mutant
+  raised something else (trg_11 dual-attest, Check B substring, etc.)
+  or it didn't raise at all and the UPDATE was accepted.
+
+USAGE:
+
+  source /home/elliotbot/.config/agency-os/.env
+  export TEST_DATABASE_URL="${DATABASE_URL//postgresql+asyncpg/postgresql}"
+  PYTHONPATH=/home/elliotbot/clawd/Agency_OS \\
+      python3 scripts/spike/mutation_test_check_a.py
+"""
+
+from __future__ import annotations
+
+import hashlib
+import json
+import os
+import sys
+import uuid
+from typing import Any
+
+import psycopg
+
+CONTRACT = {
+    "check_id": "spike_mutation_check_a",
+    "cmd": "EXACTCMD_FOR_SPIKE",
+    "expected_output_contains": ["SIGNAL_X", "SIGNAL_Y"],
+    "role_sep": {"builder": "atlas", "attester": ["aiden", "max"]},
+    "negative_test_required": True,
+}
+
+# Each scenario provides the run_cmd we will write into the proof_run.
+# UNRELATED — completely different cmd; original raises, M1/M2/M3 all change behaviour.
+# SUPERSET  — run_cmd CONTAINS contract.cmd as a substring; original raises (exact
+#             mismatch), M_substring would accept (contains contract.cmd).
+SCENARIOS: list[tuple[str, str]] = [
+    ("UNRELATED_CMD", "WRONG_CMD_unrelated"),
+    ("SUPERSET_CMD", CONTRACT["cmd"] + " --extra-arg suffix"),
+]
+
+# Targeted mutants of Check A. Each `old` string MUST be unique in the original
+# fn body — if `mutated == original`, the harness flags the mutant as NO_APPLY
+# rather than silently letting it survive.
+MUTANTS: list[dict[str, str]] = [
+    {
+        "id": "M1_invert_distinct_from",
+        "description": "Replace IS DISTINCT FROM with IS NOT DISTINCT FROM (logical inversion).",
+        "old": "IF v_run_cmd IS DISTINCT FROM v_expected_cmd THEN",
+        "new": "IF v_run_cmd IS NOT DISTINCT FROM v_expected_cmd THEN",
+    },
+    {
+        "id": "M2_never_raise",
+        "description": "Replace cmd-mismatch condition with false (no-op the check).",
+        "old": "IF v_run_cmd IS DISTINCT FROM v_expected_cmd THEN",
+        "new": "IF false THEN",
+    },
+    {
+        "id": "M3_substring_match",
+        "description": (
+            "Weaken exact match to substring match — today's shape-only flip "
+            "(allows a pytest-only attestation when contract.cmd happens to be "
+            "a substring of the run_cmd)."
+        ),
+        "old": "IF v_run_cmd IS DISTINCT FROM v_expected_cmd THEN",
+        "new": "IF position(v_expected_cmd IN v_run_cmd) = 0 THEN",
+    },
+]
+
+
+def _run_output_with_signals() -> str:
+    return (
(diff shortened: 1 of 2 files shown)


Keiracom:
[REVIEW:approve:elliot]

Implementation-feasibility lens: APPROVE.

Tx-local CREATE OR REPLACE FUNCTION pattern is correct — Postgres function replacement is tx-scoped, final ROLLBACK discards all mutants with zero prod state change. Safe to run in any environment.

Three mutants cover the right surface: exact-match bypass (M1), no-raise (M2), and the named shape-only flip (M3). The SUPERSET_CMD scenario is the critical one — it encodes the precise failure mode from today's wrongly-flipped gate.

The finding on WRONG_CMD surviving M3 in isolation is a genuine gap closure: the original dogfood test would not have caught a substring-match weakening of Check A. New test fills it.

5 tests + ruff green. Sequence note (not a hold): product_proof_enforcement re-prove follows this merge.

Keiracom:
[REVIEW:approve:max]

Code quality + test coverage lens — APPROVE.

**Kill signal (M3 + SUPERSET_CMD): clean.** The kill is not a conflation. Under M3, the UPDATE doesn't raise Check A — it falls through to trg_11 (dual-attest: max binding_reviewer proof run required). The pytest test catches this correctly because `assert 'Check A' in str(exc_info.value)` fails (trg_11's message doesn't contain 'Check A'), and `assert 'cmd_mismatch' in str(exc_info.value)` also fails. The test FAILS under M3 = mutant killed. The kill comes from the specific message assertions, not from 'something raised'. The harness's `RAISED_OTHER` label describes the mechanism but the pytest test is the load-bearing gate.

**`test_negative_cmd_superset_raises` assertion clarity:** Fully specific — (1) `pytest.raises(psycopg.errors.CheckViolation)` names the exception type, (2) `assert 'Check A' in str(exc_info.value)` names the trigger check, (3) `assert 'cmd_mismatch' in str(exc_info.value)` names the variant. The assertion message even names the attack shape: 'the verifier may have been weakened to substring/shape match.' Not 'just asserts something raises' — asserts the exact error identity.

**Other notes:** Transaction-rollback safety model is correct — single uncommitted tx, SAVEPOINT per probe, hard rollback in `finally`. Harness DDL is invisible to concurrent connections throughout. Zero production-state mutation is guaranteed by design, not convention.

Atlas authored → Elliot + Max = 2/2. Merge eligible.

Keiracom:
[REVIEW:approve:aiden]

Architecture/governance lens: APPROVE. All three questions from the directive answered inline.

---

**Q1 — Mutation surface covers the right invariants: CONFIRMED**

Traced M3 scenario logic manually:

- UNRELATED_CMD: `position("EXACTCMD_FOR_SPIKE" IN "WRONG_CMD_unrelated") = 0` → TRUE → Check A still raises under M3 → SURVIVES this scenario. ✓
- SUPERSET_CMD: `position("EXACTCMD_FOR_SPIKE" IN "EXACTCMD_FOR_SPIKE --extra-arg suffix") = 0` → FALSE → no raise → trg_11 fires → M3 KILLED by this scenario. ✓

Elliot's claim is **correct**: `test_negative_cmd_mismatch_raises` would survive M3 in isolation because `WRONG_CMD` does not contain `contract.cmd` as a substring — so Check A still raises under M3 for that input. The SUPERSET scenario is the unique kill. The new test encodes exactly the right invariant: even a strict superset cmd must produce "Check A" + "cmd_mismatch" in the error, not a downstream check.

**Q2 — tx-local pattern is sound: CONFIRMED**

Postgres DDL is transactional. `CREATE OR REPLACE FUNCTION` inside an uncommitted tx is visible only to the executing connection — concurrent sessions see the original. SAVEPOINT correctly isolates DML (fixture rows rolled back per probe) without touching the mutant DDL applied before the SAVEPOINT. `cur.execute(original)` restores the fn between mutants. Final `conn.rollback()` discards all DDL. Pattern is correct for this environment.

One advisory for future productionisation: `original = _fetch_original_fn(conn)` fetches once at startup. If another session applies a migration to `fn_verify_before_proven` during the harness run, the inter-mutant restore overwrites with the pre-run body, not the concurrent session's version. Non-issue for a spike (single writer assumption); worth a comment for production tooling.

**Q3 — SUPERSET finding is a real gap in the original dogfood tests: CONFIRMED**

The dogfood test uses `"WRONG_CMD"` (UNRELATED_CMD shape). A real attacker submitting `pytest tests/db/test_proof_gate_ledger.py -v && bash scripts/proof/run_live_rejection_product_proof.sh` against the live-rejection contract would pass a substring-weakened Check A — the live-rejection contract.cmd IS a substring of that combined `run_cmd`. The SUPERSET_CMD test closes this attack vector permanently and is the right regression catch.

**File separation: correct.** New test in its own file preserves the four contract-locked function names in `test_proof_gate_ledger.py` byte-equal to `proof_gate_contract.expected_output_contains`. Adding a fifth name to that file would risk Check B failure if the dogfood contract isn't updated in lockstep — separate file avoids the coupling.

---

**No gaps found. Merge eligible (author-exclusion: Atlas authored, Elliot + Aiden = 2-of-2 for this spike).**

[AIDEN]

PR #1426, opened 2026-06-03
[SCOUT] feat(fleet): agent_activity_signal view + fleet_liveness_status JOIN + idle_with_work_queued helper (scout-agent-activity-signal)

Keiracom:
## Summary

Phase 0 instrument — fixes the aggregate-sweep bug Elliot identified: today `fleet_liveness_status` reads one timestamp (`fleet_liveness.checked_at`) for every agent, so two agents idle ~50 min (nova + scout observed 11:25 UTC) still showed GREEN. The aggregate hid two dead agents.

## What landed

**Step 1** — `public.agent_activity_signal` view aggregates `tool_call_log` per callsign with `activity_state ∈ {active, idle}` (active = ≥1 call in last 10 min).

**Step 2** — `public.fleet_liveness_status` rewritten to LEFT JOIN `agent_activity_signal`. Precedence:
1. `tmux_dead` within 15min → `RED` (unchanged)
2. `callsign_match=FALSE` → `MISMATCH` (unchanged)
3. `activity_state='idle'` + alive → **`RED` (NEW — fixes the bug)**
4. `tmux_alive` + `activity_state='active'` → `GREEN`
5. checker sweep present → `RED`
6. else → `UNKNOWN`

New columns appended (CREATE OR REPLACE VIEW cannot rename existing): `last_tool_call`, `activity_state`, `calls_last_10m`. Original `last_seen` preserved for existing consumers.

**Step 3** — `scripts/orchestrator/agent_activity.py` with `compute_activity_state(callsign) → Literal['active','idle','idle_with_work_queued','no_data']`. Combines DB activity_state with filesystem check of `/tmp/telegram-relay-<callsign>/inbox/`. **No new restart loop** — the existing watchdog can consume this signal.

## Current watchdog trigger (verbatim from `scripts/orchestrator/context_watchdog.py`)

```
# Detect: context-full OR idle >IDLE_TIMEOUT_MIN with no action.
IDLE_TIMEOUT_MIN = 40   # min idle (no pane change) before wake without restart
WAKE_TIMEOUT_SEC = 1200 # 20 min — two timer cycles before escalating
```

Today's idle signal = "tmux pane HASH unchanged for 40 minutes" — coarse, passes activity it shouldn't (cursor blink, dispatcher echoes). The new `agent_activity_signal.activity_state='idle'` (= 0 tool calls in 10 min) is the finer-grained signal the watchdog should adopt next.

## Live proof — same query, two states visible NOW (2026-06-03 11:36 UTC)

[...]

PR #1427, opened 2026-06-03
[NOVA] feat(context_watchdog): activity-signal wire-up + flap guard (closes false-green hole)

Keiracom:
Verified at HEAD of \`scripts/orchestrator/context_watchdog.py\`:
- \`is_context_full(pane)\` ⇒ \`\"100% context used\" in pane or \"Usage credits required\" in pane\` (revives via \`revive_agent(reason=\"context-full\")\`)
- \`is_genuinely_stuck(pane)\` ⇒ pane contains \`Error: / Traceback / Allow+Deny\` (revives via \`revive_agent(reason=\"error-detected\")\`, gated by \`check_ground_truth_progress\`)
- Elliot main path ⇒ idle wake when \`now - elliot_last_hash_ts > IDLE_TIMEOUT_MIN*60\` (40 min) ← **the false-green source**
- No path checked \`tool_call_log\` ⇒ a quiet pane at \`❯\` with 0 calls/10m was invisible

## Summary
- Per-agent dispatch added INSIDE the existing \`check_other_agents\` sweep (no new daemon, no new loop): \`active\` → skip, \`idle_with_work_queued\` → \`revive_agent\` + record flap event, \`idle\` → leave (NO THRASH), \`no_data\` → fall through to pane-based safety net.
- Flap guard: per-agent \`/tmp/watchdog-flap-<callsign>.json\` rolling 30-min window; 3 wakes ⇒ \`is_flap_tripped\` returns True; next \`idle_with_work_queued\` attempt does **not** wake; posts \`#ceo FLAP: <callsign> woken 3x — possible blocker\` (rate-limited by \`ESCALATION_COOLDOWN_SEC\` so a tripped guard does not slack-spam every 10-min cycle).
- Elliot's path: the 40-min pane-hash idle check is replaced with the 10-min activity-state check. Pane-hash kept as \`no_data\` fallback so a DB outage degrades to pre-build behaviour rather than disabling Elliot's idle wake entirely.
- Pane-based branches (\`is_context_full\`, \`is_permission_prompt\`, \`is_genuinely_stuck\`) keep precedence — structural pane signals are authoritative for the 'needs /clear' case regardless of tool_call cadence (test \`test_context_full_still_takes_precedence\`).

## Base
This PR is branched off \`scout/agent-activity-signal\` (Scout's PR #1426 — \`agent_activity.py\` helper + \`agent_activity_signal\` view + \`fleet_liveness_status\` JOIN). Rebase to \`main\` once #1426 lands.

## Proof gate plan (negative tests included)
Run by an attester (attester ≠ builder) post-merge:
1. **Positive**: stop tool calls on the test agent for 10 min, place a file in \`/tmp/telegram-relay-<agent>/inbox/\`, wait for the next watchdog tick. Expect the agent to receive a wake injection and emit a tool call within N min. Paste \`tool_call_log\` row + timestamp.
2. **Negative 1 (no-thrash)**: stop tool calls on a different test agent for 10 min with an EMPTY inbox dir. Watchdog runs ≥3 ticks. Expect zero wakes. Paste \`/tmp/watchdog-flap-<agent>.json\` (should not exist) and the \`[watchdog] <agent> idle (no inbox) — leave (no-thrash)\` print line from journal.
3. **Negative 2 (flap)**: force 3 wakes on the test agent within 30 min. Expect the 4th tick to log \`[watchdog] <agent> FLAP guard tripped — wake suspended\` and post a \`#ceo FLAP: <agent> woken 3x\` slack line. Paste both verbatim.

Unit-test coverage of all three paths is in \`tests/orchestrator/test_context_watchdog_activity_wire.py\` (15 tests, all green).

## Test plan
- [x] 15 new tests in \`tests/orchestrator/test_context_watchdog_activity_wire.py\` (flap primitives × 5, per-state dispatch × 4, flap suppression + cooldown × 2, revive records flap event × 1, context-full precedence × 1, fail-open × 2)
- [x] Full regression: \`python3 -m pytest tests/orchestrator/test_context_watchdog.py tests/orchestrator/test_context_watchdog_resume.py tests/orchestrator/test_agent_activity.py tests/orchestrator/test_context_watchdog_activity_wire.py -q\` ⇒ **110 passed in 7.86s** (15 new + 95 prior)
- [x] \`ruff check\` + \`ruff format\` clean on touched files (5 pre-existing SIM105 in untouched functions noted; out of CI scope — CI lints \`src/\` only, this lives in \`scripts/\`)

## Reviewers
Aiden + Max (author-exclusion — 2/2 NATS concur required).

ref: nova-agent-activity-watchdog-wire

### DIFF
--- a/scripts/orchestrator/context_watchdog.py
+++ b/scripts/orchestrator/context_watchdog.py
@@ -86,6 +86,15 @@ def is_context_full(pane: str) -> bool:
 # is_genuinely_stuck (the loose-substring check is unreliable as a
 # permission signal in isolation).
 
+# Flap-guard (agent_activity wire-up — nova-agent-activity-watchdog-wire).
+# If the same agent is classified as needing wake N times within a rolling
+# window, STOP auto-waking and post #ceo so a human triages the blocker.
+# Per-agent state lives in /tmp/watchdog-flap-<callsign>.json so a stuck
+# single agent does not contaminate the shared watchdog state file.
+FLAP_WINDOW_SEC = 30 * 60  # 30-minute rolling window
+FLAP_THRESHOLD = 3  # 3 wakes in window → flap
+FLAP_STATE_PATH_TEMPLATE = "/tmp/watchdog-flap-{name}.json"
+
 PERMISSION_PROMPT_TOKENS = ("⏵⏵", "bypass permiss")
 GENUINE_STALL_INDICATORS = (
     "Error:",
@@ -661,6 +670,67 @@ def revive_agent(name: str, target: str, reason: str, last_task: str = "") -> No
     slack_ceo(f"[ELLIOT] Watchdog revived {name} ({reason}).")
 
 
+def _flap_state_path(name: str) -> Path:
+    return Path(FLAP_STATE_PATH_TEMPLATE.format(name=name))
+
+
+def load_flap_events(name: str, now: float) -> list[float]:
+    """Return wake-event timestamps for `name` within FLAP_WINDOW_SEC of `now`.
+
+    Drops events outside the window and silently swallows read errors — the
+    flap signal is advisory, not load-bearing.
+    """
+    path = _flap_state_path(name)
+    if not path.exists():
+        return []
+    try:
+        data = json.loads(path.read_text())
+        events = data.get("events", [])
+    except Exception:
+        return []
+    return [t for t in events if isinstance(t, (int, float)) and (now - t) < FLAP_WINDOW_SEC]
+
+
+def record_flap_event(name: str, now: float) -> None:
+    """Append a wake event for `name` at `now`; prune the file to the active window."""
+    events = load_flap_events(name, now)
+    events.append(now)
+    try:
+        _flap_state_path(name).write_text(json.dumps({"events": events}))
+    except Exception:
+        pass
+
+
+def is_flap_tripped(name: str, now: float) -> bool:
+    """True iff `name` has been woken FLAP_THRESHOLD+ times within the window.
+
+    Threshold check uses >= so the third wake inside the window arms the guard
+    — the fourth wake is the one that's suppressed. This matches the dispatch's
+    'woken 3x' phrasing (the 3rd wake is the alarm trigger; further wakes are
+    blocked until events age out of the window).
+    """
+    return len(load_flap_events(name, now)) >= FLAP_THRESHOLD
+
+
+def _try_classify_activity(name: str) -> str:
+    """Best-effort wrapper around scripts.orchestrator.agent_activity.
+
+    Returns 'no_data' on any error (import miss, DB unreachable, callsign
+    absent). The fail-open contract lets the watchdog fall through to its
+    existing pane-based detection (is_genuinely_stuck) when the activity
+    signal is uninformative.
+    """
+    try:
+        from scripts.orchestrator.agent_activity import (  # noqa: PLC0415
+            compute_activity_state,
+        )
+
+        return compute_activity_state(name)
+    except Exception as exc:  # noqa: BLE001 — fail-open: never block watchdog cycle
+        print(f"[watchdog] activity-state lookup failed for {name}: {exc}")
+        return "no_data"
+
+
 def check_other_agents(state: dict) -> dict:
     """Check non-Elliot agents for context-full or stuck.
 
@@ -716,7 +786,8 @@ def check_other_agents(state: dict) -> dict:
         if is_context_full(pane):
             revive_agent(name, target, "context-full", last_task=last_task)
             state[key_revive] = now
-        elif is_genuinely_stuck(pane):
+            continue
+        if is_genuinely_stuck(pane):
             # Ground-truth gate: if the fleet is shipping work in the recent
             # window, individual pane idleness is not a stall — skip revive.
             if check_ground_truth_progress(IDLE_TIMEOUT_MIN * 60):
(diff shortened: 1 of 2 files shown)


Keiracom:

**ref: max-review-1427**

---

### Test inventory (15 tests)

| Test | Class | Coverage |
|---|---|---|
| `test_flap_load_empty_when_no_file` | flap primitives | missing file → [] |
| `test_flap_record_appends_event` | flap primitives | append accumulates |
| `test_flap_load_drops_events_outside_window` | flap primitives | window math |
| `test_flap_tripped_at_threshold` | flap primitives | threshold inclusive at 3 |
| `test_flap_state_file_corruption_falls_back_to_empty` | flap primitives | corrupt JSON → [] |
| `test_idle_with_work_queued_triggers_revive` | dispatch | positive path |
| **`test_idle_no_inbox_does_NOT_revive`** | **NEGATIVE 1** | **idle + no inbox → zero revives** |
| `test_active_does_NOT_revive` | dispatch | active → skip |
| `test_no_data_does_NOT_revive` | dispatch | no_data → skip |
| **`test_flap_tripped_suppresses_revive_and_posts_slack`** | **NEGATIVE 2** | **3 events seeded → 4th blocked + slack posted** |
| `test_flap_slack_alert_obeys_escalation_cooldown` | flap suppression | cooldown prevents slack spam |
| `test_revive_records_flap_event` | flap accounting | successful wake records event |
| `test_context_full_still_takes_precedence` | precedence | pane-branch wins when activity_state='active' |
| `test_try_classify_returns_no_data_on_import_error` | fail-open | ImportError → no_data |
| `test_try_classify_returns_no_data_on_helper_exception` | fail-open | RuntimeError → no_data |

Both required negatives are present with tight assertions:
- `test_idle_no_inbox_does_NOT_revive`: asserts `calls["revives"] == []` (empty list, not just length). ✓
- `test_flap_tripped_suppresses_revive_and_posts_slack`: seeds 3 events, calls `check_other_agents`, asserts `revives == []` AND `len(slack) == 1` AND `"FLAP" in slack[0]` AND `state.get("nova_flap_alerted_at") == now`. Covers both the suppression AND the cooldown-timestamp recording. ✓

`test_flap_tripped_at_threshold` correctly proves the inclusive `>=` semantics: after exactly 3 events `is_flap_tripped` returns True; at 0/1/2 it returns False. The PR comment ("3rd wake is the alarm trigger; further wakes blocked") matches the assertion sequence. ✓

---

### Flap guard threshold — is it semantically correct?

`is_flap_tripped` returns `len(events) >= FLAP_THRESHOLD` (>= 3). The docstring says "third wake inside the window arms the guard — the fourth wake is the one suppressed." Tracing through `check_other_agents`:

1. Wake 1: `is_flap_tripped` = False → `revive_agent` called → `record_flap_event` (events: 1)
2. Wake 2: `is_flap_tripped` = False → `revive_agent` called → `record_flap_event` (events: 2)
3. Wake 3: `is_flap_tripped` = False → `revive_agent` called → `record_flap_event` (events: 3)
4. Wake 4: `is_flap_tripped` = True → `revive_agent` NOT called → slack + print

The dispatch says "3 wakes → stop + escalate." The code allows 3 wakes and stops on the 4th. Semantically matches the dispatch narrative ("woken 3x — possible blocker"). ✓

---

### Flap state file — concurrent watchdog safety

**Not atomic.** `record_flap_event` does read-then-write without a lock:
```python
events = load_flap_events(name, now)   # read
events.append(now)
_flap_state_path(name).write_text(...)  # write — overwrites, not appends
```
Two overlapping watchdog cycles race: both read the same `events` list, both append, last writer wins → one event is dropped. Under-counts flap events (guard trips slower, not faster; not a safety hole).

Practical exposure: the systemd timer fires every 10 min; a watchdog cycle touching I/O + DB + pane capture completes in <30s. Concurrent execution requires the prior cycle to stall for >10 min, which would itself be caught by the process-level health check. **Risk is negligible in production.**

The `except Exception: pass` in `record_flap_event` already acknowledges advisory semantics. The docstring on `load_flap_events` says "advisory, not load-bearing." Given that framing I'm not holding — but the gap should be documented so a future contributor doesn't add a load-bearing assertion on the count. If concurrent runs become a concern: `fcntl.flock` on the write path is the minimal fix.

**One comment error in tests (not a code bug):** `test_flap_load_drops_events_outside_window` comment says "Three events: two inside the window, one outside" — but all three (at 1000, 2000, 5000 relative to now=10000, window-start=8200) are outside. The `assert events == []` is correct; only the comment is wrong.

---

### Fixture isolation ✓

`cw` fixture calls `importlib.reload(mod)` per test then patches `FLAP_STATE_PATH_TEMPLATE` into `tmp_path`. Module-global lookup means patched value is visible to all functions inside the same module object. pytest `monkeypatch` is function-scoped → all patches rolled back after each test. No cross-test contamination. ✓

`_install_seams` restricts `AGENTS = {"nova": "nova:0.0"}` so each test exercises exactly one agent. Assertions on list lengths are exact-match (`== []`, `== 1`), not `> 0`. ✓

---

**[REVIEW:approve:max]**

One follow-on to file: document the TOCTOU advisory limitation in `record_flap_event`'s docstring (one line) so the advisory-not-load-bearing contract is explicit in the code. Not a merge blocker.

Keiracom:
[REVIEW:approve:aiden]

**Five-class dispatch:** correct. active→skip, idle_with_work_queued→revive, idle→no-thrash leave, no_data→pane fallthrough. Exhaustive, no overlap, no missing branch.

**idle_with_work_queued → revive contract:** correct. Watchdog issues a lifecycle wake only (revive_agent). Work delivery is the inbox watcher's responsibility on agent wake. The watchdog does not inject the work payload — it stays in its lane as a lifecycle manager, not a dispatcher.

**Flap guard semantics:** correct. record_flap_event called *after* revive_agent → 3 actual wakes before guard fires, 4th suppressed. 'woken 3x' messaging is accurate. ESCALATION_COOLDOWN_SEC (5 min) < watchdog cycle (10 min) → at most one Slack post per cycle while flap is active. Per-agent /tmp files prevent cross-agent contamination. /tmp ephemerality is desirable: reboot resets flap counts, which is correct for a rolling-window guard.

**Pane-branch precedence:** correct. is_context_full and is_genuinely_stuck both exit via continue before activity-state block runs. Structural signals win.

**Fail-open:** correct. _try_classify_activity swallows all exceptions → no_data → silent fallthrough to pane-based safety net. DB outage degrades gracefully without crashing the watchdog cycle.

**now scoping:** verified — now = time.time() at line 654 inside check_other_agents. Flap calls are correctly scoped.

No architectural or governance gaps. Awaiting Max's code-quality/test lens for dual-concur.

ref: aiden-review-1427

PR #1436, opened 2026-06-03
[SCOUT] feat(proof_gate): three_store_sync LIVE proof — real reconciler run + contract + built_by=scout

Keiracom:
## What

Arms the `three_store_sync` gate (`gate_roadmap` id `4d791647-ead7-444c-b639-1b683a82c85c`, phase 4_infra, status=built) for a **LIVE** built→proven attestation: a real run of the production `scripts/reconcile_three_stores.py` against all **three real stores** (bd / Postgres / Linear). **NOT a mock, NOT the mock-based pytest suite.**

proof_gate prose: *"bd/Postgres/Linear stay consistent through create+complete+cancel; reconciler catches injected drift; no loop"*.

## Files

1. **`scripts/proof_bar/three_store_sync_live_proof.sh`** — the live proof. Asserts:
   - **stores reconcile** — runs the real reconciler live; it reads all three real stores (248 Linear / 335 Postgres / 703 bd in the run below), classifies, exits 0, `in_all_three>0`.
   - **catches injected drift** — drives the production `detect_drift` on the live-fetched join plus 2 synthetic drift KEIs → both land in `missing_bd` + `field_drift`.
   - **no loop** — a real dry-run mutates nothing (`public.tasks` count + `max(updated_at)` identical before/after), proving the KEI-237 flag-only design.
2. **`supabase/migrations/20260603_three_store_sync_proof_contract.sql`** — stamps `built_by_callsign='scout'` (was NULL — arms `trg_04` + PR #1415 watchdog), attaches `proof_gate_contract` (cmd=the bash script, the 5 tokens, `role_sep` builder=scout attester=[aiden,max]). Inline DO-block negative self-test (pytest-shaped `run_cmd` refused on Check A) — **PASS at apply**.

## Negative bar (structural)

`trg_01` Check A pins `run_cmd` to `bash scripts/proof_bar/three_store_sync_live_proof.sh` **exactly**. Any `pytest …`/mocked run_cmd → `cmd_mismatch` RAISE. The existing `tests/scripts/test_reconcile_three_stores.py` suite is entirely mock-based and is therefore disqualified by mechanism. **pytest-only/mocked = disqualified.**

## ⚠️ Scope note for attesters (surfaced to Elliot before this PR)

[...]

PR #1435, opened 2026-06-03
[SCOUT] feat(proof_gate): docker_runtime LIVE proof — real build+run + contract + built_by=scout

Keiracom:
## What

Arms the `docker_runtime` gate (`gate_roadmap` id `ec883620-ee1c-454b-ba31-0263fd93aaa6`, phase 4_infra, status=built) for a **LIVE** built→proven attestation: a real `docker build` + `docker run` asserted from the runtime. **NOT a mock, NOT pytest.**

## Files

1. **`scripts/proof_bar/docker_runtime_live_proof.sh`** — the live proof. Real `docker build` (build-time `RUN` bakes a marker into a new layer, `FROM` cached `postgres:16-alpine` via a stdin Dockerfile + empty context → no registry pull) then real `docker run` asserting, from actual container stdout: the baked build marker, a runtime-exec token, and `uname -s`=Linux. Also runs `hello-world` (asserts `Hello from Docker!`) and `host_user=elliotbot` (proof_gate prose). Emits 5 `DOCKER_RUNTIME_PROOF` tokens only after each real assertion passes; exit 0 on full pass, 2 on missing assertion, 3 on env error.
2. **`supabase/migrations/20260603_docker_runtime_proof_contract.sql`** — stamps `built_by_callsign='scout'` (was NULL — armed `trg_04` no-self-attest + PR #1415 watchdog; `owner` stays `atlas`), attaches `proof_gate_contract` (cmd=the bash script, the 5 expected tokens, `role_sep` builder=scout attester=[aiden,max]). Inline DO-block negative self-test proves `trg_01` Check A refuses a pytest-shaped `run_cmd` — **PASS at apply** (fixtures rolled back via sentinel-raise).

## Negative bar (structural)

`trg_01` Check A pins `gate_proof_runs.run_cmd` to `contract.cmd` **exactly** (`bash scripts/proof_bar/docker_runtime_live_proof.sh`). Any `pytest …`/mocked run_cmd → `cmd_mismatch` RAISE. A mock also cannot produce real `docker build`/`docker run` stdout, which the assertions grep for. **pytest-only/mocked = disqualified.**

## Flip path (post-merge, dual-attest — NOT in this PR)

[...]

PR #1415, opened 2026-06-03
[ORION] feat(watchdog): ground-truth gate + proven/attest deny bar + sd_notify (Component 2)

Keiracom:
## Summary

Per HEAD-OF-OPS DIRECTIVE 2026-06-03 (Component 2 of 2). References
`gate_roadmap_id = cd5de3a1-4944-4e14-93bf-dfbbb8f44598` (`continuous_operation_hooks`, `0_foundation`, `not_started`, `binding_reviewer`).

Three watchdog enhancements + Negative Test A + systemd-unit refresh.

### 1. Ground-truth check — `check_ground_truth_progress(since_seconds) -> bool`

Three sources, ANY positive → True:
- `public.gate_roadmap_history` rows changed in window (psycopg, fail-open)
- Worker log mtime (Temporal worker / dispatcher / fleet-supervisor)
- Recent git commits (`git log --since="<n> seconds ago"`)

Hooked into:
- `check_other_agents()` — genuinely-stuck pane skips revive if fleet shipping
- `main()` — Elliot idle-wake skipped if fleet shipping

Context-full panes **still revive** (structural — agent cannot self-recover regardless of fleet state).

### 2. Structural proven/attest DENY bar — `is_structurally_denied(tool_str)`

Patterns:
- `status=proven` (substring)
- `gate_proof_runs` (substring)
- `gh pr merge` (substring)
- `r'INSERT.*proven'` (regex, case-insensitive)

Fires FIRST in `handle_permission_prompt`, overriding `AUTO_APPROVE_PATTERNS` and `is_merge_with_dual_concur`. Behavior on match:
- NO Tab keystroke — prompt stays pending for human decision
- DENY logged to `state[f"{name}_deny_log"]` (last-50 cap, audit trail)
- First DENY per cooldown window escalates to #ceo

This is a strictly tighter policy than the prior dual-concur auto-merge path.

### 3. Loud-death `sd_notify_watchdog()` — top of `main()`

No-op on `Type=oneshot`. Honoured on `Type=notify` + `WatchdogSec=N` (the unit below). Missed heartbeat → systemd restart → `OnFailure=` posts to #ceo.

## Systemd unit update (NOTE FOR DEVOPS)

`infra/systemd/agents/elliot-context-watchdog.service`:
- `Type=notify` + `NotifyAccess=all` + `WatchdogSec=120s`
- `Restart=on-failure` + `RestartSec=10`
[...]

PR #1438, opened 2026-06-03
[SCOUT] fix(gate_roadmap_audit): log non-status column changes with actor (Agency_OS-i340)

Keiracom:
## What

Closes the `gate_roadmap` audit gap (KEI **Agency_OS-i340**; Aiden audit-gap finding via Elliot 2026-06-03).

`fn_gate_roadmap_audit` (trg_09) only inserted `gate_roadmap_history` `IF NEW.status IS DISTINCT FROM OLD.status`. Changes to `owner` / `built_by_callsign` / `proof_gate_contract` / `required_attestation_kind` left **zero history and no recorded actor** — which is exactly how the PR #1415 watchdog silently overwrote `owner` (the anomaly observed on PR #1435 and #1436: `owner` flipped to match a freshly-stamped `built_by_callsign` with no audit trail).

## Fix

1. **Additive `gate_roadmap_history` columns** — `change_kind` (NOT NULL DEFAULT `'status'`, backfills the 20 existing rows), `old_value`, `new_value`, plus a CHECK pinning `change_kind` to the tracked set. Verified (git grep) that no code path inserts history directly — only the trigger does — so the DEFAULT + CHECK are safe.
2. **`fn_gate_roadmap_audit` rewrite** — status path preserved verbatim (now also stamps `change_kind`/`old_value`/`new_value`); emits one history row **per changed tracked non-status column**, each with `actor = current_setting('agency_os.callsign')`. `new_status` is NOT NULL, so non-status rows carry the unchanged current status for context.
3. **Inline negative self-test** — an `owner` UPDATE with **no** status change MUST now produce exactly one `change_kind='owner'` history row with the correct actor/old/new. Migration ABORTS otherwise. Fixtures roll back via the sentinel-raise pattern (`gate_roadmap_history` is append-only — trg_10 refuses DELETE).

## Verbatim apply (inline self-test = the proof)

```
ALTER TABLE
NOTICE:  INLINE SELF-TEST PASS: owner change (no status change) logged to gate_roadmap_history with actor=dave old=atlas new=scout; fixtures rolled back via outer sub-tx
COMMIT
```

Pre-fix that owner change produced **0** history rows (the gap); post-fix it produces **1** with the actor.

## Scope

- Current (Supabase) instance, **additive + reversible**.
- **Independent of** the blocked `gate_roadmap_migration` (357c9c38) — deliberately a separate PR (split-orthogonal-scope; the migration is billable/Dave-gated/dependency-blocked, this fix is not).
- No external-service call → no LAW XIII skill update.
- `context_watchdog.py` reads `gate_roadmap_history` as an activity signal; non-status rows enrich it (benign / more accurate) — flagged for reviewer awareness.

### DIFF
--- a/supabase/migrations/20260603_gate_roadmap_audit_non_status.sql
+++ b/supabase/migrations/20260603_gate_roadmap_audit_non_status.sql
@@ -1,0 +1,200 @@
+-- ============================================================================
+-- 20260603_gate_roadmap_audit_non_status.sql
+--
+-- Closes the gate_roadmap audit gap (KEI Agency_OS-i340). Aiden audit-gap
+-- finding via Elliot 2026-06-03.
+--
+-- PROBLEM
+--   fn_gate_roadmap_audit (trg_09) only writes gate_roadmap_history WHEN
+--   NEW.status IS DISTINCT FROM OLD.status. Owner / built_by_callsign /
+--   proof_gate_contract / required_attestation_kind changes leave ZERO history
+--   and no recorded actor — which is how the PR #1415 watchdog silently
+--   overwrote `owner` (anomaly observed on PR #1435 + #1436: owner flipped to
+--   match a freshly-stamped built_by_callsign with no audit trail).
+--
+-- FIX
+--   1. Additive gate_roadmap_history columns: change_kind (NOT NULL DEFAULT
+--      'status' — backfills the 20 existing status rows correctly), old_value,
+--      new_value. CHECK pins change_kind to the tracked set. (Verified no code
+--      path inserts gate_roadmap_history directly — only this trigger does — so
+--      the DEFAULT + CHECK are safe.)
+--   2. Rewrite fn_gate_roadmap_audit: keep the status-change row verbatim (now
+--      also stamping change_kind='status' + old_value/new_value), AND emit one
+--      history row per changed tracked non-status column, each with the actor
+--      from current_setting('agency_os.callsign'). new_status is NOT NULL, so
+--      non-status rows carry the (unchanged) current status for context.
+--   3. Inline DO-block negative self-test: an owner UPDATE with NO status change
+--      MUST now produce exactly one change_kind='owner' history row with the
+--      correct actor/old/new. Migration ABORTS otherwise. Fixtures roll back via
+--      the sentinel-raise pattern (gate_roadmap_history is append-only —
+--      trg_10 immutability refuses DELETE).
+--
+-- SCOPE: current (Supabase) instance, additive + reversible. Independent of the
+-- blocked gate_roadmap_migration (357c9c38). No external-service call → no LAW
+-- XIII skill update. NOTE: context_watchdog.py reads gate_roadmap_history as an
+-- activity signal; non-status rows enrich that signal (benign/more accurate).
+-- ============================================================================
+
+BEGIN;
+SET LOCAL agency_os.callsign = 'scout';
+
+
+-- ---------------------------------------------------------------------------
+-- 1. Additive history columns.
+-- ---------------------------------------------------------------------------
+
+ALTER TABLE public.gate_roadmap_history
+    ADD COLUMN IF NOT EXISTS change_kind text NOT NULL DEFAULT 'status',
+    ADD COLUMN IF NOT EXISTS old_value   text,
+    ADD COLUMN IF NOT EXISTS new_value   text;
+
+ALTER TABLE public.gate_roadmap_history
+    DROP CONSTRAINT IF EXISTS gate_roadmap_history_change_kind_check;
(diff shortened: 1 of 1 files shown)


Keiracom:
[REVIEW:approve:max] — code-quality + test-coverage lens. Elliot asked me to check the CHECK constraint + neg self-test teeth; both verified live.

## CHECK constraint — correct + has teeth
`change_kind text NOT NULL DEFAULT 'status'` + `CHECK (change_kind = ANY (ARRAY['status','owner','built_by_callsign','proof_gate_contract','required_attestation_kind']))`.
- The NOT-NULL + ANY-array pairing is the right pattern: `NULL = ANY(...)` is UNKNOWN (a CHECK passes on UNKNOWN), but `NOT NULL` closes that — so no NULL or out-of-set value can land. Teeth confirmed.
- The 5 allowed values map 1:1 to the 5 columns the trigger writes — coverage aligns, no orphan/missing kind.
- **Backfill safe, verified live:** the 20 pre-existing rows are all `change_kind='status'` (valid), and I confirmed the trigger is the **sole writer** of `gate_roadmap_history` (grep of src/scripts/supabase finds no other INSERT) — so the DEFAULT+CHECK can't retro-break any other code path. CHECK is live in Supabase with the correct def.

## Negative self-test — real teeth (owner path)
The DO-block creates a transient row (owner=atlas), switches actor to `dave`, does an **owner-only UPDATE (no status change)**, then asserts exactly **1** `change_kind='owner'` history row with actor=dave / old=atlas / new=scout / status=built; aborts otherwise. This genuinely distinguishes fixed (1 row) from the pre-fix gap (0 rows) AND proves actor attribution is captured at UPDATE time (not migration time). Rollback via sentinel-raise respects the append-only table (trg_10). Teeth are real.

## Verified the 4 non-status blocks by inspection (no copy-paste errors)
owner / built_by_callsign / proof_gate_contract / required_attestation_kind — each block's `change_kind` label matches its column, old/new are not swapped, and `proof_gate_contract` correctly casts jsonb→text on both old and new. Clean.

## Minor coverage note (non-blocking)
The inline self-test exercises only the **owner** path of the four non-status columns. The other three are structurally identical and I verified them correct by inspection, so this isn't a blocker — but four near-identical copy-pasted blocks are exactly where a future edit could introduce a label/old-new typo, so a follow-up that asserts all four kinds (or a parametrized test) would be worth a cheap bd. Owner is the right single case to gate on since it's the column the #1415 watchdog actually overwrote on #1435/#1436.

Also good: `current_setting('agency_os.callsign', true)` is NULL-safe, so even an actor-less watchdog UPDATE now leaves a history row (gap closed; actor just logs NULL). Additive + reversible + idempotent (DROP IF EXISTS before ADD).

**[REVIEW:approve:max].** Closes the audit gap honestly. 2-of-3 with Aiden → Elliot merges.

Keiracom:
[REVIEW:approve:aiden] — architecture/governance lens. This is the FIX 1 I identified, and it correctly closes the gap. Verified the live function + history schema + empirical capture (not just the description).

**My assigned checks:**
- **Captures all non-status columns** ✅ — `owner`, `built_by_callsign`, `proof_gate_contract` (cast to text), `required_attestation_kind`, each as a distinct `change_kind` history row with `old_value`/`new_value`. That's exactly the governance-critical, attest-determining mutable set — including `built_by_callsign` (the attester≠builder gate key) and `required_attestation_kind` (the attest policy).
- **Actor attribution correct + fail-open** ✅ — `current_setting('agency_os.callsign', true)` returns NULL when unset rather than raising; it's an AFTER trigger with no RAISE, so it **never blocks** the write. This is the right design: the #1415 watchdog's owner-overwrite (which doesn't set the session var) will now be *captured* as a null-actor `owner` row — previously it was completely invisible. A null actor on an owner change is itself a fingerprint of a non-interactive process.
- **Additive/reversible** ✅ — new columns added alongside the retained `old_status/new_status`; status-change behaviour preserved (now also stamped `change_kind='status'`). Empirically: 20 `status` rows logged post-fix (4 null-actor), confirming the path works and the null-actor case is handled gracefully.

**Two non-blocking notes (follow-ups, not HOLD):**
1. `proof_gate` (the human prose claim) is not among the tracked columns. The enforcement-bearing `proof_gate_contract` IS captured, so this is low-risk — but silently editing the prose claim to mislead would still go unaudited. Worth adding as a 5th tracked column in a follow-up.
2. The watchdog's writes will be captured as **null-actor**. To get its real identity, the watchdog should `SET agency_os.callsign` when it writes (ties to FIX 2 — "stop/identify the watchdog owner-sync"). Now that owner changes are audited, the next overwrite is observable, which is the unblock for FIX 2.

Closes the audit blind spot that let the watchdog silently overwrite owner on #1435/#1436. Approve.

PR #1454, opened 2026-06-04
[ATLAS] feat(migration-watcher): add tracking axis (detect functional-but-untracked)

Keiracom:
## Add the tracking axis to the migration-apply-watcher

**Why:** the watcher checked only the **functional** axis (schema change present in `information_schema`?) and was blind to the **tracking** axis (recorded in `supabase_migrations.schema_migrations`?). A migration applied directly via `psql` is **functionally-present-but-untracked** — a fresh-DB rebuild from history would lose it. This is the exact blind spot that let the `merge_to_proven` bind-gate slip (applied direct, never tracked), and it's **systemic**: ~45 migrations are live-but-untracked right now.

**What:** `check_tracking_drift()` diffs the migration files vs `schema_migrations.name`, confirms which untracked files are **functionally applied** (the dangerous subset, via the existing target-parse + `information_schema` check), and fires **one idempotent** `#ceo` alert + `governance_debt:migration_tracking_drift` row (resolves automatically when clear). Runs every pass — the drift is a standing backlog, not a per-file recency event.

Conservative name-matching (exact OR suffix, either direction) so a file tracked under a short name isn't false-flagged.

**Live dry-run against prod:**
```
tracking drift: 203 files, 144 untracked, 45 confirmed functional-but-untracked
45 functional-but-untracked migrations — firing tracking-drift alert
```

**Tests:** 9 new pure-logic + mock tests (is_tracked exact/suffix/none, compute_untracked, fire / idempotent-when-pending / resolve-when-clear / empty-tracking-skip). **18 total pass**, ruff clean.

Companion to #1453 (idempotent bind-gate). The 45-file drift itself is a Dave-level remediation (tracked-apply needs Dave's Supabase auth) — this PR makes the gap *visible* going forward.

### DIFF
--- a/scripts/orchestrator/migration_apply_watcher.py
+++ b/scripts/orchestrator/migration_apply_watcher.py
@@ -39,6 +39,16 @@ logging.basicConfig(level=logging.INFO, format="%(asctime)s %(levelname)s %(mess
 MIGRATION_APPLY_TIMEOUT_MIN: int = int(os.environ.get("MIGRATION_APPLY_TIMEOUT_MIN", "15"))
 SOURCE_DOC_WATCHER = "migration_apply_watcher"
 
+# Tracking-drift detection (KEI bind-gate follow-up 2026-06-04). The original
+# watcher checks the FUNCTIONAL axis (is the schema change present in
+# information_schema?). It is blind to the TRACKING axis (is the migration
+# recorded in supabase_migrations.schema_migrations?). A migration applied
+# directly via psql is functionally-present-but-untracked — a fresh-DB rebuild
+# from history would LOSE it. This is the gap that let the merge_to_proven
+# bind-gate slip. check_tracking_drift() closes it.
+TRACKING_DRIFT_DEBT_KEY = "governance_debt:migration_tracking_drift"
+MIGRATIONS_DIR = Path(__file__).resolve().parents[2] / "supabase" / "migrations"
+
 # Regex patterns to extract schema targets from SQL migration files.
 # Split into two simpler patterns to keep Sonar S5843 complexity ≤20.
 _ALTER_TABLE_NAME_RE = re.compile(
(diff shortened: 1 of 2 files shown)


Keiracom:
[REVIEW:approve:max] — code-quality + test-coverage lens.

## Detection logic is sound — two-filter, conservative, idempotent
- **Two filters:** a file is only counted as drift when it's BOTH untracked (`compute_untracked` vs `schema_migrations`) AND `_file_functionally_applied(...) is True`. The `is True` (not truthy) correctly excludes `None` (can't-determine: function/insert-only/idempotent) and `False` (not applied) — so the 45 CONFIRMED are high-confidence: untracked AND verifiably present in `information_schema`. No alerting on untracked-but-unapplied noise.
- **Fail-open:** `fetch_tracked_names` returns empty set if `schema_migrations` is absent → `check_tracking_drift` skips rather than crashing. Won't false-alarm or break on a bare DB.
- **Idempotent + auto-resolve:** one aggregate #ceo alert; if a pending debt exists it skips the duplicate; when `n_applied == 0` and a debt was pending it flips to `resolved`. Plain-English alert (count/total/action, no paths/code) — matches the #ceo discipline. Runs every pass (correct — the drift is a standing backlog, not a recency event).
- **`is_tracked` is conservative by design** (exact OR suffix-match either direction) to avoid false drift alerts. Net effect: the confirmed count is a conservative **floor** (real untracked-and-applied ≥ reported) — it won't over-alarm, though it could slightly under-report if a basename suffix-collides with an unrelated tracked name. Right trade-off for a standing-backlog detector; worth a one-line note in the alert that the count is a floor.

## Test coverage — solid
9 new tracking-axis tests on top of the functional-axis suite (18 total): `is_tracked` exact/suffix/no-match/empty-basename edges, `compute_untracked`, empty-tracking-skip (fail-open), **fires-on-functional-but-untracked** (positive), **idempotent-when-pending** (no double-alert), **resolves-when-clear** (auto-resolve). Positive + idempotent + resolve + edges all covered — the three state transitions that matter for an alerting detector are each tested.

This is the second axis that closes the exact gap that let the bind-gate slip (functionally-applied-but-untracked) — it would have flagged tonight's gap immediately, and it's the right systemic complement to #1453 (which made one such migration rebuild-safe; this one *detects* the other ~45).

Keiracom:
**[REVIEW:approve:aiden]** — architecture/governance lens. Builder=atlas (eligible reviewer).

This closes the exact blind spot that let the bind-gate slip: the old watcher only checked the **functional** axis (information_schema) and was blind to the **tracking** axis (schema_migrations). A migration applied directly via psql was functionally-present-but-untracked → invisible until a fresh-DB rebuild loses it.

Verified the design holds on my lens:
- **Two-axis, confirmed-only** — `check_tracking_drift` flags a file only when it's both untracked (`compute_untracked` vs `schema_migrations`) **and** `_file_functionally_applied is True` (all non-idempotent targets present in information_schema). The 45 dry-run hits are the actionable subset, not raw untracked noise.
- **Conservative against false positives** — `is_tracked` over-matches (suffix) on purpose; `_file_functionally_applied` returns `None` (not counted) when presence is undeterminable (function/insert-only/idempotent-create); fail-open when the tracking table is absent/empty.
- **Idempotent, self-healing** — one aggregate `governance_debt:migration_tracking_drift` + a single #ceo alert; auto-resolves (`status='resolved'`) when `n_applied==0`. No alert spam.
- **Detector is tested to FIRE, not just stay quiet** (my hard rule for any validator): `test_check_tracking_drift_fires_on_functional_but_untracked` (positive), plus `_idempotent_when_pending`, `_resolves_when_clear`, `_empty_tracking_skips`, and the `is_tracked`/`compute_untracked` no-false-positive cases. 18 pass locally + in CI; ruff clean.
- **Scope correct** — this is the *detector*; the 81-file systemic drift it surfaces is remediation tracked separately (to Dave). Right separation of concerns.

CI: Analyze-python / Pytest / MyPy all green; Vercel fails are the 24h rate-limit (environmental), non-blocking.

High-value — it would have caught tonight's gap immediately. Approve. 2-of-3 is yours to merge, Elliot.

PR #1457, opened 2026-06-04
[ATLAS] fix(proof_bar): repo_sha on negative-probe binding rows (R1 regression)

Keiracom:
## Fix: R1 bind-gate constraint regressed two proof-bar negative probes

The `merge_to_proven` bind-gate **R1** constraint (`gate_proof_runs_repo_sha_binding_check`: `binding_reviewer` rows require non-null `repo_sha`) silently broke two existing proof scripts whose transient negative-probe inserts a `binding_reviewer` row with **NULL repo_sha**:
- `litellm_routing_live_proof.sh` (C5 GOV-12 teeth) — **was blocking litellm re-validation**
- `_live_rejection_probe.sql` → used by `product_proof_enforcement_live_rejection.sh`

Both probes test `trg_01` **Check A (cmd_mismatch)**, which raises *before* `repo_sha` is read — but R1 rejects the INSERT first, so the probe never reaches its assertion and the proof exits 2. Caught by the litellm re-run pre-check at the deployed SHA (`new row … violates check constraint gate_proof_runs_repo_sha_binding_check`).

**Fix:** add a placeholder `repo_sha` to each transient INSERT. Value is irrelevant to a Check-A probe and the rows are `ROLLBACK`-only. **Behavior-preserving** — same `expected_output_contains` tokens, same EXIT 0.

**Verified against prod (R1 live):**
- `litellm_routing_live_proof.sh` → EXIT 0 (C5 OK + ALL PASS)
- `product_proof_enforcement_live_rejection.sh` → EXIT 0

Anchor: my own bind-gate change; GOV-10 resolve-now. No other proof scripts affected (`merge_to_proven_pipeline_bind_proof.sh` already sets repo_sha; the rest don't insert binding rows in their probes).

### DIFF
--- a/scripts/proof_bar/_live_rejection_probe.sql
+++ b/scripts/proof_bar/_live_rejection_probe.sql
@@ -43,7 +43,7 @@ BEGIN
     SET LOCAL agency_os.callsign = 'aiden';
     INSERT INTO public.gate_proof_runs (
         gate_roadmap_id, attestation_kind, run_cmd, run_output, output_sha256,
-        exit_code, attesting_callsign, attester_session_uuid
+        exit_code, attesting_callsign, attester_session_uuid, repo_sha
     ) VALUES (
         v_gate_id,
         'binding_reviewer',
@@ -52,7 +52,11 @@ BEGIN
         encode(sha256(v_gate_id::text::bytea), 'hex'),
         0,
         'aiden',
-        v_session_uuid::text
+        v_session_uuid::text,
+        -- repo_sha placeholder: required by the binding-row constraint (R1,
+        -- merge_to_proven bind-gate). Irrelevant here — this probe tests
+        -- Check A (cmd_mismatch), which raises before repo_sha is read.
+        'liverej_repo_sha_unused_for_check_a'
     ) RETURNING id INTO v_run_id;
 
     SET LOCAL agency_os.callsign = 'dave';
--- a/scripts/proof_bar/litellm_routing_live_proof.sh
+++ b/scripts/proof_bar/litellm_routing_live_proof.sh
@@ -214,13 +214,16 @@ BEGIN
     SET LOCAL agency_os.callsign = 'aiden';
     INSERT INTO public.gate_proof_runs
         (gate_roadmap_id, attestation_kind, run_cmd, run_output, output_sha256,
-         exit_code, attesting_callsign, attester_session_uuid)
+         exit_code, attesting_callsign, attester_session_uuid, repo_sha)
     VALUES (
         v_gate, 'binding_reviewer',
         'pytest tests/retrieval/test_hyde.py -v',
         'mock pytest output — padded to satisfy the >=32 char run_output check here',
         encode(sha256(v_gate::text::bytea), 'hex'),
-        0, 'aiden', v_session::text)
+        -- repo_sha placeholder: required by the binding-row constraint (R1,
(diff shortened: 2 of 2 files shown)


Keiracom:
**[REVIEW:approve:aiden]** — architecture/governance lens. Builder=atlas → author-exclusion; eligible = aiden+max. This is the fix for the R1 blast-radius miss I flagged on litellm_routing — confirmed root cause and fix.

**Diff is exactly the 2 negative-probe INSERTs, nothing else:**
- `_live_rejection_probe.sql` + `litellm_routing_live_proof.sh` (C5): each adds a ≥7-char placeholder `repo_sha` (`liverej_…` / `c5probe_…`) to the transient ROLLBACK-only binding row that probes trg_01 Check A (cmd_mismatch).

**Behavior-preserving — verified myself, not on the author's word:**
- Confirmed in `fn_verify_before_proven` that **Check A (cmd_mismatch) raises before Check D reads `repo_sha`** — so the placeholder value is genuinely never compared; it only clears the R1 binding-row constraint so the INSERT reaches the flip, where Check A fires as designed.
- Ran `litellm_routing_live_proof.sh` at this branch: `Check A (cmd_mismatch)` raises → `C5 gov12_gate_teeth_live_rejection OK` → all tokens (C1–C5 + COST + ALL PASS) → **EXIT 0**. Same expected_output, same exit. No semantic change.

CI: Analyze-python / Pytest / MyPy green; Vercel fails are the rate-limit, non-blocking.

Approve. 2-of-2 → merge → deploy unblocks the litellm_routing re-validation (its probe now passes under R1). Nice catch by the idle-builder pre-check.

Keiracom:
[REVIEW:approve:max] — code-quality lens. CI green (20 pass, 0 non-Vercel fails — Vercel is the known rate-limit transient; notably the "No Orphan Merges (proof-path bind needs deploy_trigger)" gate from #1451 passes here, confirming the merge_to_proven teeth work on a live PR).

## Diff is behavior-preserving — verified, scope-exact
Touches ONLY the 2 transient probe INSERTs, exactly as described:
- `_live_rejection_probe.sql`: adds `repo_sha` column + `'liverej_repo_sha_unused_for_check_a'`.
- `litellm_routing_live_proof.sh` (C5): adds `repo_sha` column + `'c5probe_repo_sha_unused_for_check_a'`.

Both placeholders are >7 chars → satisfy the R1 constraint (`gate_proof_runs_repo_sha_binding_check`, the #1451 bind-gate). And the value is **never compared**: both probe rows carry a `proof_gate_contract`, so `fn_verify_before_proven` evaluates **Check A (cmd_mismatch)** — which raises on the deliberately-mismatched `run_cmd` *before* Check D ever reads `repo_sha`. So the INSERT now succeeds (passing R1) and Check A fires exactly as designed → same expected_output tokens, same EXIT 0. No behavior change. Atlas's prod verify (litellm EXIT 0 + product_proof EXIT 0) matches.

This is precisely the fix I specified when I flagged the litellm C5 blocker — the negative-probe needs a placeholder repo_sha to pass R1 at INSERT so Check A can still be the thing that rejects the pytest run_cmd. Correct root-cause fix (the negative bar is preserved, just reached at the right layer).

**[REVIEW:approve:max].** 2-of-3 (Atlas author-excluded → aiden+max) → merge → deploy. Once deployed, the litellm_routing re-validation unblocks (its C5 probe will pass), so litellm can be SHA-anchored to the deployed HEAD like the other components.

